Issue #36, opened 2019-07-25
Browser - create projects when right-clicking "my projects"


wonder-sk:
Right now I can only create a project when right-clicking top-level Mergin browser item. It would be good to have that action also for the "My projects" item where I looked for it but it wasn't there.

Issue #35, opened 2019-07-25
Browser - "all projects" should be the last item


wonder-sk:
The "all projects" section should be the last one as it is not going to be commonly used. It could be also renamed to "Explore" to be consistent with browser and input

Issue #34, opened 2019-07-25
Create project dialog missing layout


wonder-sk:
The dialog is missing layout so it displays like this for me:

![image](https://user-images.githubusercontent.com/193367/61883233-56a19d80-aefa-11e9-9735-57d77861316d.png)

Issue #33, opened 2019-07-25
Auth dialog fixes


wonder-sk:
I would suggest to also have the auth dialog more similar to the login form on the web:
- show mergin logo
- use placeholder text for username/password edit widgets instead of standalone labels
- "Username or email" instead of just Username

![image](https://user-images.githubusercontent.com/193367/61882867-96b45080-aef9-11e9-843f-9ed970645f01.png)

The public instance URL https://public.cloudmergin.com/ should be pre-set in the auth dialog so that people don't need to search for it. There should be also a button to reset the URL to the default if they have managed to change it for some reason. In addition I think the URL should be less prominent as users normally won't change it.... maybe we can have a checkbox "Custom Mergin server" - when checked, it would enable mergin URL configuration.

PR #38, opened 2019-07-31
Various fixes

varmar05:
Set of small fixes related to these issues:
#6 #31 #32 #33 #34 #35 #36


Mergin/plugin.py
```diff
@@ -239,7 +240,13 @@ def createChildren(self):
     def actions(self, parent):
         action_refresh = QAction(QIcon(os.path.join(icon_path, "redo-solid.svg")), "Reload", parent)
         action_refresh.triggered.connect(self.refresh)
-        return [action_refresh]
+        actions = [action_refresh]
+
+        action_create = QAction(QIcon(os.path.join(icon_path, "plus-square-solid.svg")), "Create new project", parent)
+        action_create.triggered.connect(self.parent().create_project)
```
wonder-sk: Let's create `action_create` only when it's actually needed (for "my projects")

Mergin/ui/ui_create_project.ui
```diff
@@ -6,104 +6,109 @@
...
+        <enum>Qt::Horizontal</enum>
+       </property>
+       <property name="standardButtons">
+        <set>QDialogButtonBox::Cancel|QDialogButtonBox::Ok</set>
+       </property>
+      </widget>
+     </item>
+    </layout>
+   </item>
+  </layout>
```
wonder-sk: It would be good to add a short explanatory text to the dialog so that users understand that if they do not specify project directory, an empty project will be created. If they specify it, project data will be uploaded.

Or maybe there could be a checkbox e.g. "Also upload data to the project" and only when checked, they could pick project directory - that could be self-explaining...?

varmar05: hm, I would rather go with checkbox than explanations

saberraz: How about a drop-down option or radio buttons? 

- Create an empty folder
- This folder already contains my projects and data

varmar05: yep, that's an option, radio buttons would leave users with no doubts

wonder-sk: +1 for radio buttons

PR #57, opened 2019-10-09
storing credentials with auth manager optional 

### LINKED ISSUE #48: Endless loop with QGIS Auth password manager in a fresh profile
After mergin plugin installation, when user tries to log into to Mergin, it keeps asking for QGIS master password 100s of times. Either QGIS password manager is broken or the way Mergin is using it....

It ends up freezing QGIS.

varmar05:
Because users were bit confused with QGIS auth manager we can make it optional. They can now decide not to store credentials and once token is expired they will need to login (configure) again. In case they want use QGIS auth manager, they will be prompted for master password if not loaded automatically. 

closes #48


saberraz:
@varmar05  :+1: 

It is much clearer to the user what to do now. I have tried and wiped the password in the password manager but even if I don't tick the box for remembering password, it still remebers me in the next log in....which is not a big deal.

Auth Manager is a total mess!

varmar05:
@saberraz it can still login user even if password is not remembered in auth manager because auth is based on bearer token which is still valid. After it expires then user has to login in again using config dialog. We can remove it but I thought even if user does not want to use auth manager this prevents him to put credentials with every QGIS restart (at least for couple of hours).

PR #82, opened 2019-12-10
added method get_plugin_version and pass this version to merginClient…

PachKosti:
… constructor

relates to task https://github.com/lutraconsulting/mergin-py-client/issues/31

### DIFF
--- a/Mergin/utils.py
+++ b/Mergin/utils.py
@@ -76,7 +76,7 @@ def create_mergin_client():
     settings = QSettings()
     auth_token = settings.value('Mergin/auth_token', None)
     if auth_token:
-        mc = MerginClient(url, auth_token)
+        mc = MerginClient(url, auth_token, None, None, get_plugin_version())
         # check token expiration
         delta = mc._auth_session['expire'] - datetime.now(timezone.utc)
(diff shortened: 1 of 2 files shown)


Mergin/utils.py
```diff
@@ -99,3 +99,11 @@ def changes_from_metadata(metadata):
     updated = ", ".join(f['path'] for f in metadata['updated'])
     renamed = ", ".join(f"{f['path']} -> {f['new_path']}" for f in metadata['renamed'])
     return added, removed, updated, renamed
+
+
+def get_plugin_version():
+    with open(os.path.join(os.path.dirname(__file__),"metadata.txt"), 'r') as f:
+        for line in f:
```
varmar05: using python config parser seems more elegant to me https://docs.python.org/3.6/library/configparser.html

Mergin/utils.py
```diff
@@ -99,3 +99,11 @@ def changes_from_metadata(metadata):
...
+            if "version=" in line:
+                line = line.replace("\n", "")
+                return line.replace("version=", "")
```
varmar05: should not we return something like `plugin/version`?

Issue #81, opened 2019-12-05
When uploading from local folder


saberraz:
initialise the project folder to be use as Mergin local folder. The workflow is confusing now:

1- Create a new project using local files
2- Upload the project
3- Download the project to another location

saberraz:
A regression (tested in 2020.2)

saberraz:
See also #122

saberraz:
Fixed

PR #83, opened 2019-12-13
added username to Qgis.ini file

### LINKED ISSUE #69: Allow overwrite of directory on download
If a new Mergin project is created from a local folder then during the download project process you really want to replace the local folder. Unfortunately you get a warning that the folder already exists. There should be an option to overwrite so you don't end up with duplicates locally.

Workflow at the moment means you need to..

- Upload from local directory
- Delete local directory
- Download from Mergin

PachKosti:
because QGisAuthManager and masterpassword didn't work as we expected
Is it necessarily save Configuration again to fix be done
(closes #69 )


varmar05:
also fixes #81

PR #87, opened 2019-12-17
modified message where blank project was created

### LINKED ISSUE #40: It is not clear where the blank project is created
It will be good to make it clear that Blank project is created on Mergin Server.

PachKosti:
(closes #40)


Mergin/create_project_dialog.py
```diff
@@ -56,7 +56,7 @@ def create_project(self):
             mc.create_project(project_name, project_dir, self.ui.is_public.isChecked())
             QApplication.restoreOverrideCursor()
             settings.setValue('Mergin/localProjects/{}/path'.format(os.path.join(username, project_name)), project_dir)
-            msg = "Mergin project created successfully"
+            msg = "Mergin project created successfully" if project_dir is not None else "Blank Mergin project was created on Mergin Server"
```
varmar05: this is post submit information, but can we also inform user before? I think the best place will be to add some note in create project dialog next to 'blank project' radio button

PR #94, opened 2020-01-07
added logging errors from pyclient if were lost

### LINKED ISSUE #91: Add logging
Would be good to preserve information about errors coming from py-client<--->server communication. Usually we just show some generic message to user (what is good) but info about the actual error is lost.

PachKosti:
(closes #91)

### DIFF
--- a/Mergin/configuration_dialog.py
+++ b/Mergin/configuration_dialog.py
@@ -84,7 +84,8 @@ class ConfigurationDialog(QDialog):
                 mc = MerginClient(url, None, username, password, get_plugin_version())
                 settings.setValue('Mergin/auth_token', mc._auth_session['token'])
                 settings.setValue('Mergin/server', url)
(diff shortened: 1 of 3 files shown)


Mergin/configuration_dialog.py
```diff
@@ -97,9 +98,11 @@ def test_connection(self):
         try:
             mc = MerginClient(url, None, username, password, get_plugin_version())
             msg = "<font color=green> OK </font>"
-        except (URLError, ValueError):
+        except (URLError, ValueError) as e:
+            QgsApplication.messageLog().logMessage(str(e))
```
varmar05: I get this 'WARNING    unknown url type: '/v1/auth/login'' maybe would be good to add source of log for all entries, like 'WARNING Mergin plugin: unknown url type: '/v1/auth/login''

PR #100, opened 2020-01-13
fix error message during syncing

### LINKED ISSUE #99: Error message when syncing
I am using ~~dev instance of Mergin and~~ the master plugin and when I try to sync, I get the following error message:
image

The subsequent sync comes up with the no error message. The log file shows the sync has completed successfully.

PachKosti:
(closes #99)

### DIFF
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -253,6 +253,7 @@ class MerginProjectItem(QgsDataItem):
             
             self.mc.push_project(self.path)
 
+            qgis_files = find_qgis_files(self.path)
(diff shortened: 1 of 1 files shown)


saberraz:
Do we want to refresh the screen or reload and/or reload the survey layer or reload the whole project? Reloading the project might be a bit confusing for the user, as it zooms/pans to the project view and user gets lost in the process.

varmar05:
@saberraz we can do reload of only modified layers but it would be unnecessary complication at this stage. Let's create separate issue for it.

Issue #93, opened 2020-01-06
Problem with syncing when adding/removing columns


saberraz:
This is mainly a geodiff issue/limitation (https://github.com/lutraconsulting/geodiff/issues/16), but the plugin should handle it better.

To reproduce the error:
1- Open your project
2- Add a new column to your survey layer
3- Try to sync:
![image](https://user-images.githubusercontent.com/2649047/71804007-2ee26400-305a-11ea-9aa8-cde6a9930468.png)

saberraz:
For the time being, it should follow the old behaviour (i.e. pre geodiff): overwrites the file on server and creates a _conflict file.

saberraz:
![image](https://user-images.githubusercontent.com/2649047/73944352-54c58780-48ea-11ea-837d-112d7811615d.png)

varmar05:
@saberraz I think we can close this one?

PR #103, opened 2020-01-23
added info to status message if some files will be replaced during th…

PachKosti:
…e sync

(relates #93)

### DIFF
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -211,11 +211,7 @@ class MerginProjectItem(QgsDataItem):
             if push_renamed:
                 push_msg += f"renamed: {push_renamed} \n"
 
-            files_to_replace = ""
-            for file in push_changes["updated"]:
-                if "diff" not in file and ".gpkg" in file['path']:
-                    files_to_replace += f"{file['path']}, "
-            files_to_replace.rstrip(", ")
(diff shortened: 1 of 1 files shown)


Mergin/plugin.py
```diff
@@ -211,12 +211,19 @@ def project_status(self):
             if push_renamed:
                 push_msg += f"renamed: {push_renamed} \n"
 
+            files_to_replace = ""
+            for file in push_changes["updated"]:
+                if "diff" not in file and ".gpkg" in file['path']:
+                    files_to_replace += f"{file['path']}, "
+            files_to_replace.rstrip(", ")
```
wonder-sk: Better to build an array of strings and then do `", ".join(['hello', 'world'])` to join the items into a single string...

Mergin/plugin.py
```diff
@@ -211,12 +211,19 @@ def project_status(self):
...
+
             msg = ''
             if sum(len(v) for v in pull_changes.values()):
                 msg += pull_msg + "\n"
             if sum(len(v) for v in push_changes.values()):
                 msg += push_msg
                 msg += pretty_summary(push_changes_summary)
+                msg += f"\nNext files will be replaced on server: {files_to_replace}" if files_to_replace else ""
```
wonder-sk: 1. Better not to show this message at all if there are no files to be replaced
2. I would suggest to rephrase: "Unable to compare some of the modified local files with their server version - we will have to upload the whole file (history of the files will be lost):"

PR #105, opened 2020-01-24
Update metadata.txt

Saijin-Naib:
Fix spelling error(s)


wonder-sk:
Thanks :-)

Saijin-Naib:
Your plugin has been bit by this issue, BTW. It shows as always needing to be updated.  
https://github.com/qgis/QGIS/issues/34019

wonder-sk:
We have found that out already - hopefully a new version without trailing zero will be out next week...

PR #109, opened 2020-01-29
added function which check if is token expired. In case of yes will l…

### LINKED ISSUE #108: Handle expired token gracefully
After not using the plugin for some time, the token expires (not sure what is the exact expiration time). And with expired token things don't work as they should - for example, trying to upload/download a project would trigger a message box complaining that the project does not exist...

We should handle these cases nicely: maybe use an (existing?) endpoint to check whether we still have valid token - and if not we should try to get a new one before doing download/upload etc.

PachKosti:
…ogin user again.

(closes #108)

### DIFF
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -23,7 +23,7 @@ from urllib.error import URLError
 
 from .configuration_dialog import ConfigurationDialog
 from .create_project_dialog import CreateProjectDialog
-from .utils import find_qgis_files, create_mergin_client, ClientError, InvalidProject, changes_from_metadata
+from .utils import find_qgis_files, create_mergin_client, ClientError, InvalidProject, changes_from_metadata, LoginError
 
 icon_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "images/FA_icons")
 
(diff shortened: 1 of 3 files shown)


Mergin/plugin.py
```diff
@@ -108,6 +108,10 @@ def download(self):
             msg = "Failed to download your project {}.\n" \
                   "Please make sure your Mergin settings are correct".format(self.project_name)
             QMessageBox.critical(None, 'Project download', msg, QMessageBox.Close)
+        except LoginError as e:
+            QgsApplication.messageLog().logMessage(f"Mergin plugin: {str(e)}")
+            msg = f"<font color=red>Security token has been expired, try to renew failed, {str(e)} </font>"
```
varmar05: is not LoginError raised only if automatic token renewal failed? That would imply that probably username/password is incorrect or missing. So I would inform user about that (e.g. check configuration)

Mergin/plugin.py
```diff
@@ -226,6 +230,10 @@ def project_status(self):
         except (URLError, ClientError, InvalidProject) as e:
             msg = f"Failed to get status for project {self.project_name}:\n\n{str(e)}"
             QMessageBox.critical(None, 'Project status', msg, QMessageBox.Close)
+        except LoginError as e:
+            QgsApplication.messageLog().logMessage(f"Mergin plugin: {str(e)}")
+            msg = f"<font color=red>Security token has been expired, try to renew failed, {str(e)} </font>"
```
varmar05: a lot of duplicate messages (or code) like this, maybe better to refactor a bit

PR #120, opened 2020-03-05
Integrate async upload/download

### LINKED ISSUE #2: Progress bar when syncing a project

### LINKED ISSUE #27: QGIS UI is frozen during download/upload

wonder-sk:
This adds a dialog for download and synchronization that shows progress and can be cancelled. It should not happen anymore that the whole QGIS will become unresponsive.

Fixes #2
Fixes #27

Needs https://github.com/lutraconsulting/mergin-py-client/pull/44

image

### DIFF
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -312,7 +312,7 @@ class MerginProjectItem(QgsDataItem):
             if isinstance(dlg.exception, LoginError):
                 self._login_error_message(dlg.exception)
             elif isinstance(dlg.exception, ClientError):
-                QMessageBox.critical(self, "Project sync", "Client error: " + str(dlg.exception))
+                QMessageBox.critical(None, "Project sync", "Client error: " + str(dlg.exception))
             else:
                 # TODO: this should not happen normally - "known" exceptions should be recognized
                 msg = "Failed to upload changes!\n\n{}".format(str(dlg.exception))
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -23,6 +23,7 @@ from urllib.error import URLError
 
 from .configuration_dialog import ConfigurationDialog
 from .create_project_dialog import CreateProjectDialog
+from .sync_dialog import SyncDialog
 from .utils import find_qgis_files, create_mergin_client, ClientError, InvalidProject, changes_from_metadata, LoginError
 
 icon_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "images/FA_icons")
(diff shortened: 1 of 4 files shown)


Mergin/plugin.py
```diff
@@ -86,40 +87,56 @@ def _login_error_message(self, e):
...
+                QgsApplication.messageLog().logMessage("Mergin plugin: " + str(dlg.exception))
+                msg = "Failed to download your project {}.\n" \
+                      "Please make sure your Mergin settings are correct".format(self.project_name)
+                QMessageBox.critical(None, 'Project download', msg, QMessageBox.Close)
+            elif isinstance(dlg.exception, LoginError):
+                self._login_error_message(dlg.exception)
+            else:
+                msg = "Failed to download your project {}.\n" \
+                      "{}".format(self.project_name, str(dlg.exception))
+                QMessageBox.critical(None, 'Project download', msg, QMessageBox.Close)
```
varmar05: why we sometimes log to messageLog and sometimes not?

Mergin/plugin.py
```diff
@@ -243,41 +260,78 @@ def sync_project(self):
...
+            return
+
+        ## pull finished, start push
+
+        dlg = SyncDialog()
+        dlg.push_start(self.mc, self.path, self.project_name)
+
+        dlg.exec_()  # blocks until success, failure or cancellation
+
+        self._reload_project() # TODO: only reload project if we pulled a newer version
```
varmar05: we can do it based on status, we can either skip some stages if not needed

Mergin/ui/ui_sync_dialog.ui
```diff
@@ -0,0 +1,68 @@
...
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>Mergin - Synchronization</string>
+  </property>
+  <layout class="QVBoxLayout" name="verticalLayout">
+   <item>
+    <widget class="QLabel" name="labelMergin">
+     <property name="pixmap">
+      <pixmap>../images/logo_color.svg</pixmap>
```
varmar05: can we make it smaller, like in config dialog, it seems to me huge now
image

Mergin/plugin.py
wonder-sk: No idea - I have merely copied over that error handling handling code. I think we don't need to write anything to the QGIS message log. I would keep it as is for now and remove later when we have also better logging in the python client...

Mergin/plugin.py
wonder-sk: I would leave it as-is for the time being, as we will be changing the logic when working on https://github.com/lutraconsulting/mergin-py-client/issues/30

varmar05:
@wonder-sk I get this when sync fails (e.g. size limit) 
image

Mergin/plugin.py
varmar05: ok

varmar05: hm, now when I tried it without files to download, first sync window pop up and immediately disappeared so it results in effect of blinking, maybe it would be better to skip so it is not confusing. 

But not a hard requirement, we can postpone.

wonder-sk:
SyncError problem fixed in https://github.com/lutraconsulting/mergin-py-client/pull/44/commits/a2255261cb3d60313c143904216f8459ff859ab6

Issue #129, opened 2020-04-20
Problem downloading large projects


saberraz:
If I try to download a large project (1 something GB), I get this error:

```
An error has occurred while executing Python code: 

OverflowError: argument 1 overflowed: value must be in the range -2147483648 to 2147483647 
Traceback (most recent call last):
  File "/home/saber/.local/share/QGIS/QGIS3/profiles/test/python/plugins/Mergin/sync_dialog.py", line 103, in download_start_internal
    self.progress.setMaximum(self.job.total_size)
OverflowError: argument 1 overflowed: value must be in the range -2147483648 to 2147483647
```

saberraz:
After clicking on OK, it starts downloading....

PR #130, opened 2020-04-23
Update creation of a project with an initial push to use sync dialog

### LINKED ISSUE #127: Upload is not async for project creation
When I try to create and upload a project from a local folder, the new async function is not available and the plugin defaults to the old behaviour (freezing UI and no feedback on the progress).

wonder-sk:
Fixes #127 

This needs https://github.com/lutraconsulting/mergin-py-client/pull/50 to work

Updated "create project with upload" to use the new sync dialog with async push of the project. Also moved the "heavy duty" code out of the dialog .py file to plugin py file where the rest of pull/push calls are done.


wonder-sk:
I have sneaked in one small fix for #129 in the second commit :-)

PR #140, opened 2020-06-08
Better reporting of unhandled exceptions (fixes #132)

### LINKED ISSUE #132: Better dealing with unhandled exceptions
Right now we only report very cryptic error message:

image

To do:
- improve the message to show full exception
- add full exception report to the log

wonder-sk:
Mergin plugin will report full traceback, not just exception converted to string, which does not say much (see #132)

Note: this does not fix the underlying issue of #132, only improves error reporting.

### DIFF
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -88,6 +88,17 @@ class MerginProjectItem(QgsDataItem):
         msg = "<font color=red>Security token has been expired, failed to renew. Check your username and password </font>"
         QMessageBox.critical(None, 'Login failed', msg, QMessageBox.Close)
 
+    def _unhandled_exception_message(self, error_details, dialog_title, error_text):
+        msg = error_text + "<p>This should not happen, " \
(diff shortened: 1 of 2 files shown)


Mergin/plugin.py
```diff
@@ -316,8 +322,11 @@ def sync_project(self):
             elif isinstance(dlg.exception, ClientError):
                 QMessageBox.critical(None, "Project sync", "Client error: " + str(dlg.exception))
             else:
-                # TODO: this should not happen normally - "known" exceptions should be recognized
-                msg = "Failed to upload changes!\n\n{}".format(str(dlg.exception))
+                msg = "Failed to sync project {} due to an unhandled exception.\n\n" \
+                      "This should not happen, please report the problem here:\n" \
+                      "https://github.com/lutraconsulting/qgis-mergin-plugin/issues/new\n\n" \
```
PeterPetrik, line 327: I do not like it is 3 times the same text and link ,... maybe some helper function, so you do not need to maintain it on various places around?

wonder-sk: fixed in 58bf651

PR #155, opened 2020-07-30
Feature/variables

### LINKED ISSUE #111: Mergin specific variables for project
It will be good for the plugin to have its own variable which can be used in QGIS and Input:

Global variables:

```
mergin_username = "peter.petrik"
mergin_url = "https://public.cloudmergin.com"
```

Project variables:
```
mergin_project_name = "test_project"
mergin_project_owner = "lutraconsulting"
mergin_project_full_name = "lutraconsulting/test_project"
mergin_project_version = 10  [integer]
mergin_project_last_sync = "2020-03-01T13:00:00Z"  [datetime]
```

use https://qgis.org/pyqgis/3.0/core/Expression/QgsExpressionContextUtils.html#qgis.core.QgsExpressionContextUtils.setProjectVariable

 to set the variables

The variables will be updated when the active mergin project is changed, or when sync is finished, or when user is login/logout 
Screenshot 2020-07-16 at 13 17 37

sklencar:
Project related variables suppose to rewrite when a project is (re)loaded.
Global variable mergin user name is set when configuration window is accepted - if something went wrong during auth, the variable is removed. Is this a good behaviour?

closes #111

### DIFF
--- a/Mergin/configuration_dialog.py
+++ b/Mergin/configuration_dialog.py
@@ -84,7 +84,6 @@ class ConfigurationDialog(QDialog):
                 mc = MerginClient(url, None, username, password, get_plugin_version())
                 settings.setValue('Mergin/auth_token', mc._auth_session['token'])
                 settings.setValue('Mergin/server', url)
-                QgsExpressionContextUtils.setGlobalVariable('mergin_url', url)
             except (URLError, ClientError, LoginError) as e:
                 QgsApplication.messageLog().logMessage(f"Mergin plugin: {str(e)}")
                 mc = None
--- a/Mergin/configuration_dialog.py
+++ b/Mergin/configuration_dialog.py
@@ -89,6 +89,12 @@ class ConfigurationDialog(QDialog):
                 QgsApplication.messageLog().logMessage(f"Mergin plugin: {str(e)}")
                 mc = None
 
+        QgsExpressionContextUtils.setGlobalVariable('mergin_url', url)
+        if mc:
+            QgsExpressionContextUtils.setGlobalVariable('mergin_username', username)
+        else:
+            QgsExpressionContextUtils.removeGlobalVariable('mergin_username')
+
         return mc
 
     def test_connection(self):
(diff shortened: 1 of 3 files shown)


Mergin/configuration_dialog.py
```diff
@@ -88,6 +88,12 @@ def writeSettings(self):
                 QgsApplication.messageLog().logMessage(f"Mergin plugin: {str(e)}")
                 mc = None
 
+        QgsExpressionContextUtils.setGlobalVariable('mergin_url', url)
+        if mc:
+            QgsExpressionContextUtils.setGlobalVariable('mergin_username', username)
+        else:
+            QgsExpressionContextUtils.removeGlobalVariable('mergin_username')
+
```
wonder-sk, line 96: What if I just open QGIS without going to configuration dialog - are these variables going to be set?

Mergin/plugin.py
```diff
@@ -47,12 +52,55 @@ def initGui(self):
...
+        # Expecting [<namespace>, <project_name>, <path>]
+        key_parts = key.split('/')
+        if len(key_parts) > 2 and key_parts[-1] == 'path':
+            path = settings.value(key)
+            if path in QgsProject.instance().absoluteFilePath():
+                try:
+                    mp = MerginProject(path)
+                    metadata = mp.metadata
+
+                    if len(key_parts) > 1:
```
wonder-sk, line 81: we already tested that len(key_parts) > 2 ... why test again?

Mergin/plugin.py
```diff
@@ -47,12 +52,55 @@ def initGui(self):
...
+def set_project_variables():
+    """
+    Sets project related mergin variables. Supposed to be called when any QGIS project is (re)loaded.
+    If a loaded project is not a mergin project, it suppose to have custom variables without mergin variables by default.
+    If a loaded project is invalid - doesnt have metadata, mergin variables are removed.
+    """
+    settings = QSettings()
+    settings.beginGroup('Mergin/localProjects/')
+    for key in settings.allKeys():
+        # Expecting [<namespace>, <project_name>, <path>]
```
wonder-sk, line 72: I am a bit confused what a key really contains ... from the comment I would expect a list with three items, but it seems it is a string and I don't understand the `key_parts[-1] == 'path'` below ... could you please comment in the code what could be a valid input `key` ?

wonder-sk (approved):

PR #159, opened 2020-08-03
Added checking of the user writing permissions for the project.

DamnBack:


Mergin/plugin.py
```diff
@@ -278,6 +292,13 @@ def sync_project(self):
             QMessageBox.information(None, 'Project sync', 'Project is already up-to-date', QMessageBox.Close)
             return
 
+        # Check if user have writing rights to the project.
+        can_write = self._check_writing_permissions()
+        if can_write is False:
+            QMessageBox.information(None, "Project sync", "You have no writing rights to this project",
+                                    QMessageBox.Close)
+            return
+
```
wonder-sk, line 301: This check will also disable pull for those without write access which is not correct.

We should always let the pull to proceed and only disable push. Also, in case there are no changes to push, we do not need to show any warning to the user (and return once pull has finished).

Mergin/plugin.py
```diff
@@ -205,6 +205,16 @@ def _unsaved_changes_check(self):
...
 
+    def _check_writing_permissions(self):
+        """Check if user have writing rights to the project."""
+        info = self.mc.project_info(self.project_name)
+        username = self.mc.username()
+        writersnames = info["access"]["writersnames"]
+        if username not in writersnames:
+            return False
+        else:
+            return True
```
wonder-sk, line 216: `return username in writersnames` would be shorter :-)

Mergin/plugin.py
DamnBack: Should be fixed now.

Mergin/plugin.py
DamnBack: Done.

PR #167, opened 2020-09-30
Send project logs to developers

### LINKED ISSUE #165: Send logs to developers
It would be great if the plugin was able to submit client logs for further investigation in case of any sync issues.

varmar05:
/closes #165


Mergin/plugin.py
```diff
@@ -440,6 +441,16 @@ def remove_remote_project(self):
         except LoginError as e:
             self._login_error_message(e)
 
+    def submit_logs(self):
+        if not self.path:
+            return
+
+        log_file_name, error = send_logs(self.mc.username(), os.path.join(self.path, '.mergin', 'client-log.txt'))
```
wonder-sk, line 448: I would suggest that before sending a log, we show an information message box, saying something like this:
```
This action will send a diagnostic log to the developers.
Use this option when you encounter synchronization issues, as the log is
very useful to determine the exact cause of the problem.
It would be useful if you also send a mail to info@lutraconsulting.co.uk
and briefly describe the problem to add more context to the diagnostic log.

The log does not contain any of your data, only file names. The log file to be sent is located here:
c:\path\to\project\.mergin\client-log.txt

Please click OK if you want to proceed.

[OK]  [Cancel]
```

Mergin/utils.py
```diff
@@ -129,9 +133,56 @@ def get_qgis_version_str():
...
+
+def send_logs(username, logfile):
+    """ Send mergin-client logs to dedicated server
+
+    :param logfile: path to logfile
+    :returns: name of submitted file, error message
+    """
+    mergin_url, _, _ = get_mergin_auth()
+    system = platform.system()
+    plugin = plugin_version()
```
wonder-sk, line 156: can we rename `plugin` to e.g. `version` ?

Mergin/utils.py
```diff
@@ -129,9 +133,56 @@ def get_qgis_version_str():
...
+
+    params = {
+        "app": "plugin-{}-{}".format(plugin, system),
```
wonder-sk, line 159: from input the order is reversed: first system, then app version

Mergin/utils.py
```diff
@@ -129,9 +133,56 @@ def get_qgis_version_str():
...
+
+
+def send_logs(username, logfile):
+    """ Send mergin-client logs to dedicated server
+
+    :param logfile: path to logfile
+    :returns: name of submitted file, error message
+    """
+    mergin_url, _, _ = get_mergin_auth()
+    system = platform.system()
```
wonder-sk, line 155: can we have system in lowercase, so that it's more consistent with input (also using lowercase for system)

Mergin/images/FA_icons/bars-solid.svg
```diff
@@ -0,0 +1 @@
+<svg aria-hidden="true" focusable="false" data-prefix="fas" data-icon="bars" class="svg-inline--fa fa-bars fa-w-14" role="img" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 448 512"><path fill="currentColor" d="M16 132h416c8.837 0 16-7.163 16-16V76c0-8.837-7.163-16-16-16H16C7.163 60 0 67.163 0 76v40c0 8.837 7.163 16 16 16zm0 160h416c8.837 0 16-7.163 16-16v-40c0-8.837-7.163-16-16-16H16c-8.837 0-16 7.163-16 16v40c0 8.837 7.163 16 16 16zm0 160h416c8.837 0 16-7.163 16-16v-40c0-8.837-7.163-16-16-16H16c-8.837 0-16 7.163-16 16v40c0 8.837 7.163 16 16 16z"></path></svg>
```
wonder-sk, line 1: What do you think about using e.g. medkit icon (https://fontawesome.com/icons/medkit?style=solid) instead of the generic hamburger icon (typically used to open another menu)

wonder-sk (approved):

PR #176, opened 2020-11-04
Add validation of URL

### LINKED ISSUE #171: Custom mergin URL validation
If a user checks 'Custom Mergin server' in configuration dialog, but leaves the URL empty, a python error will be raised when closing the dialog, and then also the browser dock integration stays broken.

We should do some basic sanity check - e.g. whether the URL starts with `http://` or `https://`- and maybe not accept invalid URL at all...

sklencar:
closes #171

### DIFF
--- a/Mergin/utils.py
+++ b/Mergin/utils.py
@@ -1,5 +1,6 @@
 import os
 from datetime import datetime, timezone
+import json
 import urllib.parse
 import urllib.request
 from urllib.error import URLError, HTTPError
@@ -186,3 +187,17 @@ def send_logs(username, logfile):
         return log_file_name, None
     except (HTTPError, URLError) as e:
         return None, str(e)
+
+
+def validate_mergin_url(url):
+    url_ping = urllib.parse.urljoin(url, urllib.parse.quote("/ping"))
+    try:
+        with urllib.request.urlopen(url_ping) as f:
+            response = json.load(f)
+            if response.get('service') != 'Mergin':
+                return 'URL is invalid Mergin URl.'
+            if response.get('status') != 'online':
+                return 'Mergin server is offline.'
+    except:
+        return "Invalid URL"
+    return None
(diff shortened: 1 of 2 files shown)


Mergin/utils.py
```diff
@@ -186,3 +187,17 @@ def send_logs(username, logfile):
...
+def validate_mergin_url(url):
+    url_ping = urllib.parse.urljoin(url, urllib.parse.quote("/ping"))
+    try:
+        with urllib.request.urlopen(url_ping) as f:
+            response = json.load(f)
+            if response.get('service') != 'Mergin':
+                return 'URL is invalid Mergin URl.'
+            if response.get('status') != 'online':
+                return 'Mergin server is offline.'
+    except:
```
wonder-sk, line 201: Catch-all "except" is a bad practice - please specify what exception should be caught

Mergin/utils.py
```diff
@@ -186,3 +187,17 @@ def send_logs(username, logfile):
         return log_file_name, None
     except (HTTPError, URLError) as e:
         return None, str(e)
+
+
+def validate_mergin_url(url):
+    url_ping = urllib.parse.urljoin(url, urllib.parse.quote("/ping"))
```
wonder-sk, line 193: I would suggest to use mergin client's `is_server_compatible()` with a temporary MerginClient object (auth not needed) rather than doing the query directly.

And we don't need to deal with "online" status - if the server is in maintenance mode, that's okay (it's only a temporary thing).

Mergin/utils.py
```diff
@@ -186,3 +187,17 @@ def send_logs(username, logfile):
         return log_file_name, None
     except (HTTPError, URLError) as e:
         return None, str(e)
+
+
+def validate_mergin_url(url):
```
wonder-sk, line 192: please add a docstring

wonder-sk (approved):

PR #177, opened 2020-11-05
Added namespace parameter for creating a new project

### LINKED ISSUE #169: Allow project creation to be placed under other namespaces user has write access to
E.g. if a user has write access to an oranization, it should be possible to create the project under that namespace too.

sklencar:
Depends on: https://github.com/lutraconsulting/mergin-py-client/pull/69
closes #169

### DIFF
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -597,7 +597,7 @@ class MerginRootItem(QgsDataCollectionItem):
 
     def show_create_project_dialog(self):
         user_info = self.mc.user_info()
-        dlg = CreateProjectDialog(username=user_info['username'], user_organisations=user_info['organisations'])
+        dlg = CreateProjectDialog(username=user_info['username'], user_organisations=user_info.get('organisations', []))
         if not dlg.exec_():
             return  # cancelled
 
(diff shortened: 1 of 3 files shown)


Mergin/plugin.py
```diff
@@ -596,20 +596,21 @@ def configure(self):
...
-    def create_project(self, project_name, project_dir, is_public):
+    def create_project(self, project_name, project_dir, is_public, namespace):
         """ After user has selected project name, this function does the communication.
         If project_dir is None, we are creating empty project without upload.
         """
 
         QApplication.setOverrideCursor(Qt.WaitCursor)
         try:
-            self.mc.create_project(project_name, is_public)
+            self.mc.create_project(project_name, is_public, namespace)
```
wonder-sk, line 613: I can't see support for namespaces in mergin-py-client...

wonder-sk (commented):
When I click menu item to create a new project:
```
AttributeError: 'MerginClient' object has no attribute 'user_info' 
Traceback (most recent call last):
  File "/home/martin/.local/share/QGIS/QGIS3/profiles/default/python/plugins/Mergin/plugin.py", line 599, in show_create_project_dialog
    user_info = self.mc.user_info()
AttributeError: 'MerginClient' object has no attribute 'user_info'
```
Are we missing a PR for mergin client?

Also, can we make the form look nicer, something like ths:

Screenshot from 2020-11-09 17-16-30

(no mention of "namespace", in a single row, with a slash between the fields)

PR #190, opened 2020-12-15
Revise metadata if seems to not be correct

### LINKED ISSUE #187: Problem uploading data to organisation
I have an organisation, but when I try to create and upload data from the plugin, I get the following error message:
`Client error: The requested URL was not found on the server.`

sklencar:
Revise metadata if seems to not be correct after a project has been transferred (namespace has been changed).
Note that this solution fixes only mergin plugin itself, not mergin-py-client since information about the change is from listing projects. Therefore is has to be probably done manually for the console client (similarly to set remote url on git).

closes #187

### DIFF
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -160,6 +160,7 @@ class MerginProjectItem(QgsDataItem):
         
         if self.path:
             self.setIcon(QIcon(os.path.join(icon_path, "folder-solid.svg")))
+            self._revise_metadata(self.project_name, self.path)
         else:
             self.setIcon(QIcon(os.path.join(icon_path, "cloud-solid.svg")))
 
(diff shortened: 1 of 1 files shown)


Mergin/plugin.py
```diff
@@ -278,6 +279,24 @@ def _have_writing_permissions(self):
...
+    def _revise_metadata(self, project_full_name, path):
+        """
+        Validates project name and namespace. If those properties in metadata doesnt match `project_full_name`,
+        metada are updated accordingly. Such case can occur after project transfer, when metadata has not been updated yet.
+        :param project_full_name: project's <namespace>/<name>. This argument suppose to be from server
+        :param path: Absolute location of a project
+        :return bool: True if metadata has been updated.
+        """
+        mp = MerginProject(path)
+        meta_full_name = mp.metadata["name"]
```
varmar05, line 291: looking into MerginProject class and metadata can raise some exception, should not we handle this?

Mergin/plugin.py
```diff
@@ -278,6 +279,24 @@ def _have_writing_permissions(self):
...
+        if meta_full_name != project_full_name:
+            meta_copy = mp.metadata
+            meta_copy["name"] = project_full_name
+            mp.metadata = meta_copy
+            return True
```
varmar05, line 296: what is the use of this return value?

varmar05:
@sklencar let's also fix mapping in QGIS ini in case project has been downloaded before

wonder-sk:
The original ticket was fixed in a different way - closing this PR...

PR #193, opened 2021-01-15
Added plugin toolbar

erpas:


erpas:
Plugin toolbar screenshot:

mergin_plugin_toolbar

Mergin/plugin.py
```diff
@@ -623,6 +703,21 @@ def actions(self, parent):
         return actions
 
 
+class MerginProjectsManager(object):
```
DamnBack, line 706: Could you add some docstrings with brief class description?

Mergin/plugin.py
```diff
@@ -76,8 +82,68 @@ def initGui(self):
...
+    def enable_toolbar_actions(self, enable=True):
+        for action in self.toolbar.actions():
+            if action.text() == "Configure Mergin Plugin":
+                continue
+            action.setEnabled(enable)
+
+    def configure(self):
+        self.data_item_provider.root_item.configure()
+
+    def cur_project_status(self):
```
DamnBack, line 138: Can we have "current" instead of just "cur" in method name?

Mergin/plugin.py
```diff
@@ -76,8 +82,68 @@ def initGui(self):
...
+        self.data_item_provider.root_item.manager.status(self.mergin_proj_name)
+
+    def cur_project_sync(self):
```
DamnBack, line 141: Can we have "current" instead of just "cur" in method name?

Mergin/plugin.py
erpas: Done

DamnBack:
Looks good to me now. @wonder-sk feel free to take it over. :)

PR #218, opened 2021-04-12
Fetch projects using the paginated list endpoint

erpas:
Also fixed multiple project managers creation issue.

### DIFF
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -3,6 +3,7 @@
 # GPLv3 license
 # Copyright Lutra Consulting Limited
 
+from math import floor
 import sip
 import os
 import shutil
@@ -22,7 +23,7 @@ from qgis.core import (
     QgsProviderRegistry,
 )
 from qgis.utils import iface
-from qgis.PyQt.QtWidgets import QAction, QFileDialog, QMessageBox, QApplication
+from qgis.PyQt.QtWidgets import QAction, QFileDialog, QMessageBox, QApplication, QDockWidget
 from qgis.PyQt.QtCore import QSettings, Qt
 from urllib.error import URLError
 
@@ -36,15 +37,17 @@ from .utils import (
     LoginError,
     create_mergin_client,
     find_qgis_files,
-    icon_path,
     get_mergin_auth,
+    icon_path,
+    is_number,
     login_error_message,
     mergin_project_local_path,
+    PROJS_PER_PAGE,
+    remove_project_variables,
     same_dir,
     send_logs,
     unhandled_exception_message,
     unsaved_project_check,
-    remove_project_variables,
 )
 
 from .mergin.merginproject import MerginProject
@@ -59,7 +62,10 @@ class MerginPlugin:
         self.actions_always_on = []
         self.menu = "Mergin Plugin"
         self.mergin_proj_dir = None
+        self.mc = None
         self.manager = None
+        self.wizard = None
+        self.error = None
         self.toolbar = self.iface.addToolBar("Mergin Toolbar")
         self.toolbar.setToolTip("Mergin Toolbar")
 
@@ -80,12 +86,6 @@ class MerginPlugin:
         # Triggering auth request to QGIS auth framework already at this point will make sure that
         # the dialog asking for master password is started from the main thread -> no crash.
         get_mergin_auth()
-
-        self.data_item_provider = DataItemProvider()
-        QgsApplication.instance().dataItemProviderRegistry().addProvider(self.data_item_provider)
-        # related to https://github.com/lutraconsulting/qgis-mergin-plugin/issues/3
-        # if self.iface.browserModel().initialized():
-        #     self.iface.browserModel().reload()
 
         self.create_manager()
 
@@ -123,6 +123,14 @@ class MerginPlugin:
             enabled=False,
             always_on=False,
         )
+
+        self.enable_toolbar_actions()
+
+        self.data_item_provider = DataItemProvider(self)
+        QgsApplication.instance().dataItemProviderRegistry().addProvider(self.data_item_provider)
+        # related to https://github.com/lutraconsulting/qgis-mergin-plugin/issues/3
+        # if self.iface.browserModel().initialized():
+        #     self.iface.browserModel().reload()
 
     def add_action(
         self,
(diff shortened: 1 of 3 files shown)


erpas:
Currently, Mergin plugin only allows to fetch 100 projects in a group.

This PR allows for getting more projects in chunks of 50. Each projects group has the Fetch more option in the right-click menu, see below. Also the numbers of projects already downloaded and available on the server are shown.

obraz

Additionally, a "Double-click for more..." item is added at the end of the list, if there are more projects to fetch from the service.

obraz

Mergin/plugin.py
```diff
@@ -162,61 +170,87 @@ def add_action(
...
         """Open plugin configuration dialog."""
-        if self.data_item_provider.root_item is None:
-            QMessageBox.warning(
-                None,
-                "Mergin Plugin",
-                "Mergin plugin first use requires QGIS restart.",
-            )
-            return
-        self.data_item_provider.root_item.configure()
+        self.show_browser_panel()
```
wonder-sk, line 226: :thinking:  not sure if we need this... and especially to do it before opening the config dialog... I would at least move that further down when the config dialog was accepted...

Mergin/plugin.py
```diff
@@ -262,13 +294,14 @@ def unload(self):
...
 
-    def __init__(self, parent, project, mc, project_manager):
+    def __init__(self, parent, project, project_manager):
         self.project = project
         self.project_name = posixpath.join(
             project["namespace"], project["name"]
         )  # we need posix path for server API calls
         QgsDataItem.__init__(self, QgsDataItem.Collection, parent, self.project_name, "/Mergin/" + self.project_name)
         self.path = mergin_project_local_path(self.project_name)
+        self.setSortKey(f"1 {self.name()}")
```
wonder-sk, line 304: This would deserve probably a comment :-) not sure why it is there...

Mergin/plugin.py
```diff
@@ -575,71 +625,116 @@ def __init__(self, parent, grp_name, grp_filter, icon, order, project_manager):
...
+                self.removeChildItem(self.fetch_more_item)
+            except RuntimeError:
+                pass
+            self.fetch_more_item = None
+        fetched_count = self.rowCount() + len(self.projects)
+        if fetched_count < self.total_projects_count:
+            self.fetch_more_item = FetchMoreItem(self)
+            self.fetch_more_item.setState(QgsDataItem.Populated)
+            sip.transferto(self.fetch_more_item, self)
+        group_name = f"{self.group_name} ({fetched_count}/{self.total_projects_count})"
```
wonder-sk, line 683: I don't really like the idea to show "XXX (fetched/total)" for each group, sorry :-)  maybe just show "XXX (total)" but including the number of fetched projects is IMHO not needed...

Mergin/plugin.py
```diff
@@ -575,71 +625,116 @@ def __init__(self, parent, grp_name, grp_filter, icon, order, project_manager):
...
+        self.setName(group_name)
+
+    def fetch_more(self):
+        """Fetch another page of projects and add them to the group item."""
+        if self.total_projects_count - self.rowCount() <= 0:
+            QMessageBox.information(None, "Fetch Mergin Projects", "All projects already listed.")
```
wonder-sk, line 689: Can this ever happen? Do we allow people to fetch more if there's nothing else to fetch?

Mergin/plugin.py
```diff
@@ -575,71 +625,116 @@ def __init__(self, parent, grp_name, grp_filter, icon, order, project_manager):
...
         except URLError:
             error_item = QgsErrorItem(self, "Failed to get projects from server", "/Mergin/error")
             sip.transferto(error_item, self)
             return [error_item]
         except Exception as err:
             error_item = QgsErrorItem(self, "Error: {}".format(str(err)), "/Mergin/error")
             sip.transferto(error_item, self)
-            return [error_item]
+            return None, [error_item]
+        return None
```
wonder-sk, line 652: Hmm what happens with the return values? in different occasions we return `[error_item]` or `None, [error_item]` or `None` ? What are we supposed to return?

Mergin/plugin.py
erpas: It is here to:
* make sure projects are listed first ("1") before the `Fetch more` item
* get the projects ordered by namespace AND name - currently we can only ordered by a single project attr on the server - that was namespace to make fetching more project process consistent. 

When the `order_params` of the `paginated_projects_list` gets into play on the server and we will get the list already sorted, but still we need a way to put the `Fetch more...` item at the bottom of a group tree branch.

Mergin/plugin.py
erpas: I was think about users fetching more of Explore group projects, so they clearly know how many of them are already fetched - there are nearly 1K of them ATM. It is true that users have the projects ordered alphabetically and can kind of guess that there is plenty of them to be fetched if they see D as the first letter of projects name for example.
It is also true, that there won't be Fetch more... item at the bottom of the list if there is nothing more to get.
And also true that for "My Projects" and "Shared with me" projects there not that many of them and it looks a bit odd.

I will show only the total number of projects in each group, as you suggest.
How about having the complete count info in the tool tip for group items?

Mergin/plugin.py
erpas: Nope, this is a leftover from an earlier stage when the "Fetch more..." item was not removed from the list when all projects were already fetched.
Fixed.

Mergin/plugin.py
```diff
@@ -59,7 +62,10 @@ def __init__(self, iface):
         self.actions_always_on = []
         self.menu = "Mergin Plugin"
         self.mergin_proj_dir = None
+        self.mc = None
         self.manager = None
+        self.wizard = None
+        self.error = None
```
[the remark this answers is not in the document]
erpas: Yep.

Mergin/plugin.py
```diff
@@ -59,7 +62,10 @@ def __init__(self, iface):
         self.actions_always_on = []
         self.menu = "Mergin Plugin"
         self.mergin_proj_dir = None
+        self.mc = None
         self.manager = None
+        self.wizard = None
```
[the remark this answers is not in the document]
erpas: Yep.

Mergin/plugin.py
erpas: Moved down.

Mergin/plugin.py
erpas: Yep, this is a mess. We either need None (all OK) of an error item (as a list item), if something goes wrong.

wonder-sk (approved):

PR #247, opened 2021-05-19
Local Mergin projects are now browser directory items

### LINKED ISSUE #28: Being able to list layers from projects in Mergin
Currently, only the name of projects appear in the Browser panel from plugin. It will be good to be see the list of layers too.

### LINKED ISSUE #236: Mergin project always opened after sync
Sometimes you wish you sync a mergin project but NOT open it.  Nowadays, if there is a sync down or up, when the sync finishes, the mergin project is opened.  When you have another project open already which takes ages to open, this is very annoying.

erpas:
Closes #28
Closes #236 

obraz

### DIFF
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -13,17 +13,18 @@ from qgis.PyQt.QtCore import pyqtSignal
 from qgis.PyQt.QtGui import QIcon
 from qgis.core import (
     QgsApplication,
-    QgsDataItem,
     QgsDataCollectionItem,
-    QgsErrorItem,
-    QgsExpressionContextUtils,
+    QgsDataItem,
     QgsDataItemProvider,
     QgsDataProvider,
+    QgsDirectoryItem,
+    QgsErrorItem,
+    QgsExpressionContextUtils,
     QgsProject,
     QgsProviderRegistry,
 )
 from qgis.utils import iface
-from qgis.PyQt.QtWidgets import QAction, QFileDialog, QMessageBox, QApplication, QDockWidget
+from qgis.PyQt.QtWidgets import QAction, QFileDialog, QMessageBox, QDockWidget
 from qgis.PyQt.QtCore import QSettings, Qt
 from urllib.error import URLError
 
@@ -45,7 +46,6 @@ from .utils import (
     PROJS_PER_PAGE,
     remove_project_variables,
     same_dir,
-    send_logs,
     unhandled_exception_message,
     unsaved_project_check,
 )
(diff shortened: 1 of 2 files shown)


erpas:
It should also close https://github.com/lutraconsulting/qgis-mergin-plugin/issues/236

Mergin/plugin.py
```diff
@@ -298,31 +298,25 @@ def __init__(self, parent, project, project_manager):
         )  # we need posix path for server API calls
         QgsDataItem.__init__(self, QgsDataItem.Collection, parent, self.project_name, "/Mergin/" + self.project_name)
         self.path = mergin_project_local_path(self.project_name)
```
wonder-sk, line 300: if it is a remote project, it should have no local path...?

Mergin/plugin.py
```diff
@@ -298,31 +298,25 @@ def __init__(self, parent, project, project_manager):
...
+        self.setSortKey(f"1 {self.name()}")
+        self.setIcon(QIcon(icon_path("cloud-solid.svg")))
         self.project_manager = project_manager
         if self.project_manager is not None:
             self.mc = self.project_manager.mc
         else:
             self.mc = None
 
+    def open_project(self):
+        self.project_manager.open_project(self.path)
```
wonder-sk, line 310: if it is a remote project, how can we open the project...?

Mergin/plugin.py
```diff
@@ -369,6 +358,96 @@ def download(self):
...
+        try:
+            self.mc.clone_project(self.project_name, dlg.project_name, dlg.project_namespace)
+            msg = "Mergin project cloned successfully."
+            QMessageBox.information(None, "Clone project", msg, QMessageBox.Close)
+            self.parent().reload()
+            # we also need to reload My projects group as the cloned project could appear there
+            group_items = self.project_manager.get_mergin_browser_groups()
+            if "My projects" in group_items:
+                group_items["My projects"].reload()
+        except (URLError, ClientError) as e:
```
wonder-sk, line 377: I would prefer to keep the try/except blocks as short as possible, so it is clear where the error is coming from - e.g. here it should be enough to just wrap the clone_project() call?

wonder-sk (commented):
Looking good, just minor bits...

Mergin/plugin.py
```diff
@@ -367,18 +364,18 @@ def clone_remote_project(self):
...
-            self.parent().reload()
-            # we also need to reload My projects group as the cloned project could appear there
-            group_items = self.project_manager.get_mergin_browser_groups()
-            if "My projects" in group_items:
-                group_items["My projects"].reload()
         except (URLError, ClientError) as e:
             msg = "Failed to clone project {}:\n\n{}".format(self.project_name, str(e))
             QMessageBox.critical(None, "Clone project", msg, QMessageBox.Close)
         except LoginError as e:
             login_error_message(e)
```
wonder-sk, line 371: let's not forget adding `return` to both exception handlers :-)

wonder-sk (approved):

Issue #186, opened 2020-12-09
"Error: Invalid token type"


HugoP27:
Good day, 
The plugin gives the following error when I open a blank QGIS project and try to connect to Mergin:

An error has occurred while executing Python code: 

ValueError: Invalid token type 
Traceback (most recent call last):
  File "C:/Users/hugop/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\Mergin\plugin.py", line 591, in configure
    self.mc = dlg.writeSettings()
  File "C:/Users/hugop/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\Mergin\configuration_dialog.py", line 88, in writeSettings
    mc = create_mergin_client()
  File "C:/Users/hugop/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\Mergin\utils.py", line 116, in create_mergin_client
    return MerginClient(url, mc._auth_session['token'], username, password, get_plugin_version())
  File "C:/Users/hugop/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\Mergin\mergin\client.py", line 51, in __init__
    token_data = decode_token_data(auth_token)
  File "C:/Users/hugop/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\Mergin\mergin\client.py", line 28, in decode_token_data
    raise ValueError("Invalid token type")
ValueError: Invalid token type

Python version: 3.7.0 (v3.7.0:1bf9cc5093, Jun 27 2018, 04:59:51) [MSC v.1914 64 bit (AMD64)] 
QGIS version: 3.16.1-Hannover Hannover, b381a90dca 

Python Path:
C:/PROGRA~1/QGIS3~1.16/apps/qgis/./python
C:/Users/hugop/AppData/Roaming/QGIS/QGIS3\profiles\default/python
C:/Users/hugop/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins
C:/PROGRA~1/QGIS3~1.16/apps/qgis/./python/plugins
C:\PROGRA~1\QGIS3~1.16\apps\Python37
C:\PROGRA~1\QGIS3~1.16\apps\Python37\Scripts
C:\Program Files\QGIS 3.16\bin\python37.zip
C:\PROGRA~1\QGIS3~1.16\apps\Python37\DLLs
C:\PROGRA~1\QGIS3~1.16\apps\Python37\lib
[...]

PR #249, opened 2021-05-20
Add global Mergin client log to user-submitted logs

erpas:
It is a part of a fix for #186 - https://github.com/lutraconsulting/mergin-py-client/pull/91

### DIFF
--- a/Mergin/utils.py
+++ b/Mergin/utils.py
@@ -245,7 +245,7 @@ def send_logs(username, logfile):
     global_logs = b""
     if global_log_file and os.path.exists(global_log_file):
         with open(global_log_file, 'rb') as f:
-            if os.path.getsize(logfile) > 512 * 1024:
+            if os.path.getsize(logfile) > 100 * 1024:
                 f.seek(-100 * 1024, os.SEEK_END)
             global_logs = f.read() + b"\n--------------------------------\n\n"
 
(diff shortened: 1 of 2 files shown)


Mergin/utils.py
```diff
@@ -240,12 +242,17 @@ def send_logs(username, logfile):
             username
         )
 
+    global_logs = "".encode()
+    if global_log_file:
+        with open(global_log_file, 'rb') as f:
+            global_logs = f.read() + "\n--------------------------------\n\n".encode()
```
wonder-sk, line 248: we could do something similar to what is below - if the size of global log is larger than e.g. 100kb, we would only take the last 100kb...

Mergin/utils.py
```diff
@@ -242,10 +242,12 @@ def send_logs(username, logfile):
             username
         )
 
-    global_logs = "".encode()
-    if global_log_file:
+    global_logs = b""
+    if global_log_file and os.path.exists(global_log_file):
         with open(global_log_file, 'rb') as f:
-            global_logs = f.read() + "\n--------------------------------\n\n".encode()
+            if os.path.getsize(logfile) > 512 * 1024:
```
wonder-sk, line 248:
```suggestion
            if os.path.getsize(logfile) > 100 * 1024:
```

wonder-sk (approved):

PR #268, opened 2021-06-18
Reload updated layers after sync

### LINKED ISSUE #267: QGIS canvas doesn't refresh after sync
Then new features are added into project in INPUT and then sync to the QGIS, the point are not visible immediately on the map.
The map needs to be refreshed to see the features. This can be confusing for the user.

erpas:
Closes #267.

### DIFF
--- a/Mergin/projects_manager.py
+++ b/Mergin/projects_manager.py
@@ -266,11 +266,10 @@ class MerginProjectsManager(object):
         qgis_proj_filename = QgsProject.instance().fileName()
         qgis_proj_basename = os.path.basename(qgis_proj_filename)
         qgis_proj_changed = False
-        updated_sources = []
         for updated in pull_changes["updated"]:
-            updated_sources.append(updated["path"])
             if updated["path"] == qgis_proj_basename:
                 qgis_proj_changed = True
+                break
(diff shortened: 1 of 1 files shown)


Mergin/projects_manager.py
```diff
@@ -291,6 +292,10 @@ def sync_project(self, project_dir, project_name=None):
             # TODO: report success only when we have actually done anything
             msg = "Mergin project {} synchronized successfully".format(project_name)
             QMessageBox.information(None, "Project sync", msg, QMessageBox.Close)
+            # reload updated layers
+            for layer in QgsProject.instance().mapLayers().values():
+                if os.path.basename(layer.source()) in updated_sources:
+                    layer.reload()
```
wonder-sk, line 298: is it necessary to call `layer.reload()` like this? I thought it would be enough to just refresh map canvas...

erpas: Unfortunately, it is required. Simple canvas refresh doesn't help here.

erpas:
Clearing canvas cache + refresh is sufficient for having the changes visible immediately after syncing.

wonder-sk (approved):

PR #269, opened 2021-06-24
ci for packaging (part-1 no fixes for libs rpaths)

PeterPetrik:
Add CI support to generate Mergin QGIS Plugin. The pygeodiff libraries are included, but NOT patched to use QGIS sqlite library, so most probably it plugin will not load in QGIS


wonder-sk (approved):
:clap: :clap: awesome

Issue #261, opened 2021-06-08
Update packaging script for geodiff 1.0


wonder-sk:
Now that new geodiff does not statically link sqlite anymore, we need to update packaging for mergin plugin:
- linux
   ```bash
   patchelf --replace-needed libsqlite3-d9e27dab.so.0.8.6 libsqlite3.so.0.8.6 libpygeodiff-1.0.0-python.so
   patchelf --remove-rpath libpygeodiff-1.0.0-python.so
   ```
- windows - nothing needed
- mac os - ?

PeterPetrik:
`9e27dab` in the name of sqlite3 lib may be mangling on geodiff wheel side... maybe we should disable mangling at all ?


macOS: 
```
Install_name_tool -add_rpath <rpath> <library_path_on_disk> 
install_name_tool -delete_rpath <rpath> <library_path_on_disk>
 install_name_tool -change <old> <new> <library_path_on_disk>
```
list linked:
`otool -L <library_path_on_disk>`

wonder-sk:
Fixed in #269, #270, #271, #272

PR #270, opened 2021-06-30
[CI] Packaging for linux

wonder-sk:
Remove the internal sqlite3 dependency from pygeodiff on linux (sqlite3 will be used from qgis install)

Refs #261

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -28,7 +28,8 @@ jobs:
 
       - name: Patching pygeodiff binaries
         run: |
-          echo "no-op !! Here is the place to patch linkage !!"
+          patchelf --replace-needed libsqlite3-d9e27dab.so.0.8.6 libsqlite3.so.0.8.6 ./pygeodiff-binaries/libpygeodiff-1.0.0-python.so \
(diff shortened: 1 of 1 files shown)


.github/workflows/packages.yml
```diff
@@ -28,7 +28,8 @@ jobs:
 
       - name: Patching pygeodiff binaries
         run: |
-          echo "no-op !! Here is the place to patch linkage !!"
+          patchelf --replace-needed libsqlite3-d9e27dab.so.0.8.6 libsqlite3.so.0.8.6 ./pygeodiff-binaries/libpygeodiff-1.0.0-python.so
```
PeterPetrik, line 31: 1. `d9e27dab` is most probably name-mangling addition of audit wheel command. This shouldn't be hardcoded here, but rather ldd + regex should be used to get the mangled name + the version of sqlite (0.8.6) and this should be used instead
2. don't you want to link agains libsqlite3.so (without version suffix?)
3. libpygeodiff-1.0.0 version shouldn't be hardcoded, use libpygeodiff-${GEODIFF_VER}

PeterPetrik (approved):

PR #308, opened 2021-11-02
Packaging fixes

### LINKED ISSUE #273: GeoTIFFs are rewritten when packaging project
When creating a project and using the packaging option, raster layers are copied using QgsRasterFileWriter. This is fine as a solution for the most general case, but if the user has a GeoTIFF, we should handle it more efficiently without rewriting the file.

I have seen a case when a cloud-optimized GeoTIFF with 2MB size was expanded to 700+MB file as a result, because no compression was applied to it (and in addition to that no tiling, no overviews).

There are two issues to solve:
- in case of GeoTIFF (and maybe some other formats in future), it would be best to just take the file as it is and copy it to the destination folder as it is
- in more general case, apply some reasonable options (maybe use COG as the output gdal driver, to include tiling and overviews)

cc @lieti

### LINKED ISSUE #277: Create Mergin Project and unavailable layers: Value returned by tree_layer.layer() not checked for None.
Value returned by `tree_layer.layer()` is not checked for `None` in `__init__` of `LayerTreeProxyModel` class in create_project_wizard.py (line 230).

When clicking on "Create Mergin Project" toolbar button with an opened QGIS project containing unavailable layers (e.g. invalid data source), the Mergin plugin crashes with the following Python error:

```
File "/home/ondrej/.local/share/QGIS/QGIS3/profiles/default/python/plugins/Mergin/plugin.py", line 245, in create_new_project
wizard = NewMerginProjectWizard(
File "/home/ondrej/.local/share/QGIS/QGIS3/profiles/default/python/plugins/Mergin/create_project_wizard.py", line 400, in __init__
self.package_page = PackagingPage(parent=self)
File "/home/ondrej/.local/share/QGIS/QGIS3/profiles/default/python/plugins/Mergin/create_project_wizard.py", line 373, in __init__
self.layers_view = PackageLayersTreeView()
File "/home/ondrej/.local/share/QGIS/QGIS3/profiles/default/python/plugins/Mergin/create_project_wizard.py", line 353, in __init__
self.proxy_model = LayerTreeProxyModel(self)
File "/home/ondrej/.local/share/QGIS/QGIS3/profiles/default/python/plugins/Mergin/create_project_wizard.py", line 230, in __init__
lid = tree_layer.layer().id()
AttributeError: 'NoneType' object has no attribute 'id'
```

The documentation of the QgsLayerTreeLayer.layer() function at https://qgis.org/pyqgis/3.16/core/QgsLayerTreeLayer.html#qgis.core.QgsLayerTreeLayer.layer warns about this:

> [QgsLayerTreeLayer.layer()  returns the map layer associated with this node.] This can be (and often is!) None, e.g. in the case of a layer node representing a layer which has not yet been fully loaded into a project, or a layer node representing a layer with an invalid data source. The returned pointer must ALWAYS be checked to avoid dereferencing None.

If the code only needs to retrieve the layer ID string, calling just `tree_layer.layerId()` instead of `tree_layer.layer().id()` may suffice. When I fix this and try to package the project with some unavailable layers, however, the plugin crashes with:
```
  Traceback (most recent call last):
  File "/home/ondrej/.local/share/QGIS/QGIS3/profiles/default/python/plugins/Mergin/create_project_wizard.py", line 467, in accept
  layer_state = proxy_model.layers_state[layer.id()]
  AttributeError: 'NoneType' object has no attribute 'id'
  ```
So it seems that this bug may be present in multiple parts of the code.

(Mergin plugin version 2021.4, QGIS 3.16.8, Ubuntu 20.04)

### PR DESCRIPTION
**Pull request by erpas**
Fixes #273 - do not rewrite GeoTiffs 
Fixes #277 - added a check for tree node layers validity - invalid layers are kept as is by default - they might be valid layers in a specific environment only.

### DIFF
--- a/Mergin/metadata.txt
+++ b/Mergin/metadata.txt
@@ -1,7 +1,7 @@
 ; the next section is mandatory
 [general]
 name=Mergin
-qgisMinimumVersion=3.16
+qgisMinimumVersion=3.10
 qgisMaximumVersion=3.99
 description=Handle Mergin projects
 version=2021.4.2
--- a/Mergin/metadata.txt
+++ b/Mergin/metadata.txt
@@ -1,7 +1,7 @@
 ; the next section is mandatory
 [general]
 name=Mergin
-qgisMinimumVersion=3.10
+qgisMinimumVersion=3.16
 qgisMaximumVersion=3.99
 description=Handle Mergin projects
 version=2021.4.2
--- a/Mergin/utils.py
+++ b/Mergin/utils.py
@@ -569,6 +569,9 @@ def is_layer_packable(layer):
     if provider_name in QGIS_DB_PROVIDERS:
         return layer.type() == QgsMapLayerType.VectorLayer
     else:
+        if provider_name == "gdal":
+            # for GDAL rasters check it is a local file
+            return os.path.isfile(layer.dataProvider().dataSourceUri())
         return provider_name in PACKABLE_PROVIDERS
 
 
@@ -614,14 +617,15 @@ def package_layer(layer, project_dir):
         save_raster_layer(layer, project_dir)
 
     else:
-        # meshes and anything else
+        # everything else (meshes)
         raise PackagingError("Layer type not supported")
 
 
 def save_raster_layer(raster_layer, project_dir):
     """
     Save raster layer to the project directory.
-    If the source raster is a GeoTiff, create a copy of the original file in the project directory.
+    If the source raster is a local GeoTiff, create a copy of the original file in the project directory.
+    Remote COG rasters are kept as they are.
     If it is a GeoPackage raster, save it also as a GeoPackage table.
     Otherwise, save the raster as GeoTiff using Qgs with some optimisations.
     """
@@ -629,7 +633,10 @@ def save_raster_layer(raster_layer, project_dir):
     driver_name = get_raster_driver_name(raster_layer)
 
     if driver_name == "GTiff":
-        copy_tif_raster(raster_layer, project_dir)
+        # check if it is a local file
+        is_local = os.path.isfile(raster_layer.dataProvider().dataSourceUri())
+        if is_local:
+            copy_tif_raster(raster_layer, project_dir)
     elif driver_name == "GPKG":
         save_raster_to_geopackage(raster_layer, project_dir)
     else:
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: Mergin/utils.py [hunk 1]
```diff
@@ -602,59 +607,112 @@ def package_layer(layer, project_dir):
...
+    """Save raster layer as GeoTiff in the project directory."""
+    new_raster_filename = get_unique_filename(raster_layer.name() + ".tif")
+    layer_filename = os.path.join(project_dir, os.path.basename(new_raster_filename))
+    # Get data type info to set a proper compression type
+    dp = raster_layer.dataProvider()
+    is_byte_data = [dp.dataType(i) <= Qgis.Byte for i in range(raster_layer.bandCount())]
+    compression = "JPEG" if all(is_byte_data) else "LZW"
+
+    raster_writer = QgsRasterFileWriter(layer_filename)
+    raster_writer.setCreateOptions([f"COMPRESS={compression}"])
```
wonder-sk, line 679: I would suggest to also use `TILED=YES` - otherwise pixels are stored line by line which is typically less efficient (especially for larger rasters)

Mergin/metadata.txt
```diff
@@ -1,7 +1,7 @@
 ; the next section is mandatory
 [general]
 name=Mergin
-qgisMinimumVersion=3.10
+qgisMinimumVersion=3.16
```
wonder-sk, line 4: Do we need to raise the minimum version?

Mergin/utils.py
```diff
@@ -602,59 +607,112 @@ def package_layer(layer, project_dir):
...
+    """
 
-        provider_opts = QgsDataProvider.ProviderOptions()
-        provider_opts.layerName = layer.name()
-        datasource = layer_filename
-        layer.setDataSource(datasource, layer.name(), "gdal", provider_opts)
+    driver_name = get_raster_driver_name(raster_layer)
 
+    if driver_name == "GTiff":
+        copy_tif_raster(raster_layer, project_dir)
```
wonder-sk, line 632: Just wondering - does packaging work if the source raster is a COG hosted on some remote server? (we don't need to support that use case, but it should not fail with a python error - instead we should just disable packaging of remote COGs if that's not supported)

Mergin/metadata.txt
erpas: It is currently hard to test new code against now obsolete version on various platforms.

erpas: Reverted to 3.10 with the belief that it should work also for 3.10. If not - users will report issues and we bump the version.

Mergin/utils.py
erpas: Added a check for COGs - if they are remote the packaging option for them is disabled and they can be either skipped of kept as they are. In the future we can add also packaging of remote rasters (and possibly db-stored too).

wonder-sk (approved):
Looks good, thanks!

PR #311, opened 2021-11-11
Add attention banner

tomasMizera:
**PR REQUIRES NEW VERSION OF MERGIN_PY_CLIENT!**

This PR adds action required banner to plugin. 

`/v1/user/service` endpoint is used to find out if the banner should be visible based on `action_required` key. API call has been added to mergin py client.

Banner is shown inside Qgis message bar as error message without timer. It is shown each time user logs in or starts QGIS while being logged in automatically.  

Screenshots:

image (1)

image (2)

> Note: latter message grammar errors will be fixed on Mergin side

### DIFF
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -22,9 +22,11 @@ from qgis.core import (
     QgsExpressionContextUtils,
     QgsProject,
     QgsProviderRegistry,
+    QgsMessageLog,
+    Qgis
 )
 from qgis.utils import iface
-from qgis.PyQt.QtWidgets import QAction, QFileDialog, QMessageBox, QDockWidget
+from qgis.PyQt.QtWidgets import QAction, QFileDialog, QMessageBox, QDockWidget, QPushButton
 from qgis.PyQt.QtCore import QSettings, Qt
 from urllib.error import URLError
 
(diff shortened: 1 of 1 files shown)


Mergin/plugin.py
```diff
@@ -232,6 +237,24 @@ def configure(self):
             self.on_config_changed()
             self.show_browser_panel()
 
+    def post_login(self):
+        """Groups actions that needs to be done when auth information changes"""
```
varmar05, line 241: Is this only action after login so far?

Mergin/plugin.py
```diff
@@ -232,6 +237,24 @@ def configure(self):
...
+        # check action required flag
+        service_response = self.mc.user_service()
+
+        if service_response and type(service_response) is dict:
+            requires_action = service_response.get("action_required", False)
+
+            if requires_action:
+                iface.messageBar().pushMessage(
+                    "Mergin",
+                    f"Your attention is required.&nbsp;Please visit the <a href={self.mc.url}>Mergin dashboard</a>",
```
varmar05, line 254: I would rather be explicit (/dashboard) in case there would be some issue with redirect.

Do we want to use UTM tags (similar to Input)?

Mergin/plugin.py
```diff
@@ -199,6 +203,7 @@ def on_config_changed(self):
         """Called when plugin config (connection settings) were changed."""
         self.create_manager()
         self.enable_toolbar_actions()
+        self.post_login()
```
varmar05, line 206: How does it work in plugin with auto-refresh of auth bearer token (e.g. QGIS session is open for several days)?
If we have it, I would consider checking attention_required as well.

wonder-sk: Probably I would not tie auto-refresh of token to the action required check - I would suggest to keep it simple and not worry too much about this case of qgis being open for several days...

wonder-sk (approved):

PR #323, opened 2021-12-08
Add option to configure selective sync

### LINKED ISSUE #282: Configure selective sync for Input
This will need:
- addition of an extra tab in project properties throught QGIS API
- the tab will have a checkbox to enable/disable selective sync + optionally specify which sub-folder to use
- read/write this info from/to the dedicated file in mergin folder

varmar05:
This PR adds a new tab in projects properties related to mergin project. Currently it enables to configure selective sync for Input. If open project is not a mergin project, this option is disabled.

Needs update to QGIS 3.16

Closes #282

### DIFF
--- a/Mergin/project_settings_widget.py
+++ b/Mergin/project_settings_widget.py
@@ -1,0 +1,73 @@
+import json
+import os
+from qgis.PyQt import uic
+from qgis.gui import QgsOptionsWidgetFactory, QgsOptionsPageWidget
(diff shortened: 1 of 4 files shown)


Mergin/project_settings_widget.py
```diff
@@ -0,0 +1,73 @@
+import json
```
wonder-sk, line 1: Add header with license

wonder-sk (approved):

PR #342, opened 2022-03-01
Stop sync if conflicts were created and reword corresponding message (fix #330)

### LINKED ISSUE #330: Improve pull behavior when there are new conflict files
The behavior during sync when some conflict files get created should be improved:
- we should not ask user if he/she wants to continue - we should simply show a warning message and stop further sync (it is difficult for the user to decide whether Yes or No should be clicked without more information)
- the wording of the message box should be improved - now we talk about some conflicts that were found - but in fact we have just _created_ them, and we should try to point users to some docs with more details what are the conflicts about, why they appear and how to avoid them...

Relevant code here: https://github.com/lutraconsulting/qgis-mergin-plugin/blob/master/Mergin/projects_manager.py#L237

alexbruy:
Stop sync if conflicts were created and show warning with the list of conflicting copies. Reword warning message to point user to documentation.

Fixed #330.

### DIFF
--- a/Mergin/projects_manager.py
+++ b/Mergin/projects_manager.py
@@ -235,19 +235,23 @@ class MerginProjectsManager(object):
             return
 
         if dlg.pull_conflicts:
-            msg = "Following conflicts between local and server version found: \n\n"
+            msg = "Sync aborted as there are conflicting changes in local and server versions.\n" +
+                  "Following conflict files were created:\n\n"
             for item in dlg.pull_conflicts:
                 msg += item + "\n"
             msg += (
-                "\nYou may want to fix them before upload otherwise they will be uploaded as new files. "
-                "Do you wish to proceed?"
+                "\nTo learn what are the conflicts about and how to avoid them please check our"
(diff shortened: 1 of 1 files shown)


alexbruy:
As published docs does not have anything about conflicts yet (or I was not able to find this page) I use the section which looks like a possible place for this information.

Mergin/projects_manager.py
```diff
@@ -235,19 +235,23 @@ def sync_project(self, project_dir, project_name=None):
             return
 
         if dlg.pull_conflicts:
-            msg = "Following conflicts between local and server version found: \n\n"
+            msg = "Sync aborted as there are conflicting changes in local and server versions.\n" +
```
wonder-sk, line 238: Saying that sync was aborted is probably not entirely correct, as pull finished successfully, it "just" created some conflicted copies. I would suggest to use the following wording:
`
We have detected conflicting changes between your local copy and the server that could not be resolved automatically. It is recommended to manually reconcile your local changes (which have been moved to conflicted copies) before running sync again.
`

alexbruy: Done.

wonder-sk (approved):
Thanks Alex!

Issue #327, opened 2022-01-20
Sync breaks after database schema change (make_copy_sqlite error)


wonder-sk:
Steps to reproduce:
1. user A downloads a project
2. user A adds a new feature to the project (but does not sync yet)
3. user B modifies database schema (e.g. adds a column to a table in geopackage) and syncs
4. user A syncs - this will first fail to rebase (which is fine) but then also fail when creating conflict copy - it will cause an unexpected sync error:

![sync-after-schema-change](https://user-images.githubusercontent.com/193367/150357147-ab46e64a-38b2-4140-bdb0-8bfcbe47dd9c.png)

After that, due to the unexpected sync error, the project directory is left in a broken state and further syncs are not successful.

This only happens on Windows - on this system, files that are currently open can't get deleted (replaced).

The full python error:
```
Traceback (most recent call last):

  File "C:/Users/martin/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\Mergin\mergin\merginproject.py", line 437, in apply_pull_changes
    self.geodiff.create_changeset(basefile, src, server_diff)

  File "C:/Users/martin/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\Mergin\mergin\deps\pygeodiff\main.py", line 65, in create_changeset
    return self.clib.create_changeset(base, modified, changeset)

  File "C:/Users/martin/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\Mergin\mergin\deps\pygeodiff\geodifflib.py", line 200, in create_changeset
    _parse_return_code(res, "createChangeset")

  File "C:/Users/martin/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\Mergin\mergin\deps\pygeodiff\geodifflib.py", line 41, in _parse_return_code
    raise GeoDiffLibError(msg)

Mergin.mergin.deps.pygeodiff.geodifflib.GeoDiffLibError: createChangeset


During handling of the above exception, another exception occurred:


Traceback (most recent call last):

  File "C:/Users/martin/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\Mergin\sync_dialog.py", line 306, in pull_timer_tick
    self.pull_conflicts = pull_project_finalize(self.job)

[...]

PR #344, opened 2022-03-10
Handle unfinished pull

alexbruy:
Implement additional checks and logic for handling unfinished pull, as described in #327.

Following workflow was implemented:

1. When sync initiated plugin checks whether previous pull was completed successfully. If unfinished pull was found we try to finish it (if project is opened in QGIS it will be closed beforehand and then re-opened). As finishing previous pull means creation of conflicted copies, sync is stopped and user asked to check conflicts to prevent data loss. 
2. If project is in clean state first phase of sync (pull) started
3. Pull might fail and project will end up in the unfinished pull state. So we close project and try to finish pull. At this point sync will stop, because on success we will have conflict files and on failure we still have unfinished pull.
4. If pull completed without any errors we are going to push changes.

Also add a warning about unfinished pull to the status dialog if project has one
unfinished-pull-warning

And display message bar on project load if it has unfinished pull
open_warning

To provide user with the list of conflict files created while finalizing unfinished pull I have modified Python client code to provide this information lutraconsulting/mergin-py-client#129

### DIFF
--- a/Mergin/projects_manager.py
+++ b/Mergin/projects_manager.py
@@ -3,7 +3,7 @@ import os
 from qgis.core import QgsProject
 from qgis.utils import iface
 from qgis.PyQt.QtWidgets import QMessageBox, QApplication
-from qgis.PyQt.QtCore import QSettings, Qt
+from qgis.PyQt.QtCore import QSettings, Qt, QTimer
 from urllib.error import URLError
 
 from .sync_dialog import SyncDialog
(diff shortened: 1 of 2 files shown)


PeterPetrik (approved):

Mergin/projects_manager.py
```diff
@@ -51,14 +51,29 @@ def have_writing_permissions(self, project_name):
...
     def open_project(project_dir):
         if not project_dir:
             return
 
         qgis_files = find_qgis_files(project_dir)
         if len(qgis_files) == 1:
             iface.addProject(qgis_files[0])
+            if self.mc.has_unfinished_pull(project_dir):
+                widget = iface.messageBar().createMessage(
+                    "Unfinished pull",
```
wonder-sk, line 63: I would suggest to rename this title to "Mergin", so it is clear where this message is coming from

Mergin/projects_manager.py
```diff
@@ -234,26 +266,23 @@ def sync_project(self, project_dir, project_name=None):
...
+        # we stop and ask user to examine them.
+        if self.mc.has_unfinished_pull(project_dir):
+            current_project_path = os.path.normpath(QgsProject.instance().absolutePath())
+            delay = 0
+            if current_project_path == os.path.normpath(project_dir):
+                QgsProject.instance().clear()
+                delay = 2500
+            # we have to wait a bit to let the OS (Windows) release lock on the GPKG files
+            # otherwise attempt to resolve unfinished pull will fail
+            QTimer.singleShot(delay, lambda: self.resolve_unfinished_pull(project_dir, True))
```
wonder-sk, line 281: as this code to close project + deferred resolve is present three times, it may be good to move it to a separate function, so that the logic is in one place? (especially if we ever need to modify the delay time or change the logic to retry few times)

Issue #289, opened 2021-09-14
Warning in project status dialog if user has modified structure of a GeoPackage


wonder-sk:
Right now often users are not aware that changing database structure can introduce conflict files, so let's warn them in the project status dialog at the bottom, e.g. something like this:

![image](https://user-images.githubusercontent.com/193367/133262734-8c94ce0c-ff88-4631-b6b9-06dbb2c84c2e.png)

The "learn more" link would lead to the online documentation.

When user does not use "project status" dialog and only clicks "sync" button from toolbar/browser, we should still show the warning (or maybe the full status dialog?) so that this message is not lost.

PeterPetrik:
https://merginmaps.com/docs/manage/synchronisation/
https://merginmaps.com/docs/manage/plugin-sync-project/
https://merginmaps.com/docs/manage/deploy-new-project/

PeterPetrik:
@alexbruy this is already fixed?

alexbruy:
Yes, it was fixed.

PR #353, opened 2022-03-16
Add more validation checks

### LINKED ISSUE #254: Check if the Attachment widget is using the correct settings
the path is relative and warn the users if it is not.

alexbruy:
Add more validation checks:

- using relative path in the Attachment widget
- Attachment widget does not use local path, e.g. `/home/user/some/path`
- expression-based path should not be set as a default path (to work correctly it should be set as data-defined)
- database structure was changed. For now only simple check that there are no new/removed/renamed fields

Fixes #254 and #289.

ref CU-27pvwte


Mergin/utils.py
```diff
@@ -882,3 +882,19 @@ def is_number(s):
         return False
     except TypeError:
         return False
+
+
+def has_schema_change(mp, layer):
```
PeterPetrik, line 887: proper way is to add "has_schema_change" to (py)geodiff, since at the end of the day geodiff is one that needs to decide if it is for conflict or not... @wonder-sk do you think it is worth adding it now? (I think you can combine few API calls to get the info already, but it is not easiest)

wonder-sk: You can do something like this:
```
geodiff.schema('sqlite', '', '/home/martin/project/file.gpkg', '/tmp/schema_local.json')
geodiff.schema('sqlite', '', '/home/martin/project/.mergin/file.gpkg', '/tmp/schema_base.json')
# now compare the two json files
```

I agree that it would be nice to have API in geodiff to check for that directly - just raised https://github.com/lutraconsulting/geodiff/issues/164 for that - but it's probably not necessary to add it right now...

wonder-sk: Also agreed that we should use geodiff, because there are other bits that `layer.fields()` does not cover - e.g. geometry type change, or new table added etc

PeterPetrik: I think it is faster to add to geodiff. Making 2 tmp files and then deleting them(?), doing diff sounds like more work than implementation in geodiff for all platforms we support.

PeterPetrik (commented):
Happy with changes! can you please add automatic tests?

alexbruy:
Tests added and I tried to add them to the workflow, but I have never worked with GitHub actions before. Seems I'm doing something wrong when mounting volume to docker image and as a result test runner can't find tests.

PeterPetrik (approved):
👏

PR #354, opened 2022-03-16
Fix "Remove locally" functionality (fix #352)

### LINKED ISSUE #352: Can't remove project with 'remove locally' option
On QGIS LR 3.24.0 and latest plugin (2021.6) sometimes I can't remove local project

Steps to reproduce
1) log in to the mergin plugin
2) download an project and open it
3) remove it locally 
4) expected behavior is that the project is removed. But I get this error:
Capture2

Interesting thig is also, that after download, slash between user name and project name is on different side in mergin plugin browser

Capture

alexbruy:
On Windows when QGIS project is cleared OS needs some time to release lock on open files. So when we try to remove files they still might be locked. As a result removing local project copy will file. To avoid this we add a small delay before attempt to remove Mergin project files.

Fixes #352.

### DIFF
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -535,6 +535,9 @@ class MerginLocalProjectItem(QgsDirectoryItem):
                 mp.log.removeHandler(log_file_handler)
                 del mp
 
+                # as releasing lock on previously open files takes some time
+                # we have to wait a bit before removing them, otherwise rmtree
+                # will fail and removal of the local rpoject will fail as well
                 QTimer.singleShot(250, lambda: shutil.rmtree(self.path))
             except PermissionError as e:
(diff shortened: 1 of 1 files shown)


Mergin/plugin.py
```diff
@@ -535,7 +535,7 @@ def remove_local_project(self):
                 mp.log.removeHandler(log_file_handler)
                 del mp
 
-                shutil.rmtree(self.path)
+                QTimer.singleShot(250, lambda: shutil.rmtree(self.path))
```
PeterPetrik, line 538: can we add comment to code why QTimer is added?

alexbruy: Done

PeterPetrik (approved):

PR #355, opened 2022-03-17
take into account namespace permissions to get user privileges (fix #351)

### LINKED ISSUE #351: Can't write to the project of the organization
In latest version of the plugin (2021.6), user can't upload any change with error `You have no writing rights to this project`.
It is for writer, admin and owner members rights.
It is possible to upload to the project from web interface, so I believe, it is error in plugin

https://user-images.githubusercontent.com/83582181/158535630-bb6347bd-063a-4656-ad4c-cd85be9a97d4.mp4

alexbruy:
As described in #351 when user who is not an owner of the organization tries to sync project of the organization they get error. It happens because plugin relies on the `access` field of the project information to determine whether user has write permissions on a given project. However `access` field is taken from the explicit project permissions, while invited to the organization users get access via this organization.

In order to overcome this issue we need to check write access also by using `permissions` field, as this field is evaluated with respect to namespace.

Also need to fix the same issues in the Python client lutraconsulting/mergin-py-client#130

Fixes #351

### DIFF
--- a/Mergin/projects_manager.py
+++ b/Mergin/projects_manager.py
@@ -49,7 +49,12 @@ class MerginProjectsManager(object):
         info = self.mc.project_info(project_name)
         username = self.mc.username()
         writersnames = info["access"]["writersnames"]
-        return username in writersnames
+        permissions = info["permissions"]
+        # permissions field contains information about update, delete and upload
+        # privileges of the user on a specific project. This is more accurate
+        # information then "writernames" field, as it takes into account
+        # namespace privileges. So we have to check both "writersnames" and "permissions"
+        return (username in writersnames) or all(permissions.values())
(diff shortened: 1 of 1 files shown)


Mergin/projects_manager.py
```diff
@@ -49,7 +49,12 @@ def have_writing_permissions(self, project_name):
         info = self.mc.project_info(project_name)
         username = self.mc.username()
         writersnames = info["access"]["writersnames"]
-        return username in writersnames
+        permissions = info["permissions"]
```
varmar05, line 52: we do pretty much the same in py-client code, wondering if it is not better to create function only there

varmar05 (approved):

PR #367, opened 2022-03-31
Improve validation/sync workflow and status dialog overhaul

alexbruy:
Show project status dialog before sync, so users can revise changes and be aware of any possible misconfigurations/changes which may break their project. Both "Sync" and "Status" toolbar buttons and context menu item will open the Project Status dialog. When invoked via "Sync" button the Project Status dialog will have a "Sync" button to start synchronization, otherwise only "Close" button will be shown.

Improve status dialog UI by:

- saving/restoring its size and position
- dropping tabs and making validation results more visible to user. If there are no warnings, validation results widget will be hidden to save space and avoid confusion.
- using message bar at the top of the dialog instead of the plain label at the bottom to display messages
- use rich text widget for validation results. This allows us to use long descriptions for warnings and include links to the documentation. Better text and links to be added.

In addition approach to handling validation results was changed a bit. For some checks (like the one checking for online/offline layers) we show all affected layers under the warning name. While results of other checks are grouped by a layer. This should make it easier for users to review warnings and fix them.

Before (left - Project Status tab, right - Validation results tab)
зображення

After (left - Status dialog, both status and validation results are visible, right - Status dialog, validation widget is hidden as there are no warnings)
зображення

### DIFF
--- a/Mergin/help.py
+++ b/Mergin/help.py
@@ -1,0 +1,14 @@
+
+HELP_ROOT = "https://merginmaps.com/docs"
+
+
+class MerginHelp:
+    """Class for generating Mergin plugin help URLs."""
+
+    def howto_attachment_widget(self):
+        return f"{HELP_ROOT}/layer/settingup_forms/"
+
+    def howto_background_maps(self):
+        return f"{HELP_ROOT}/gis/settingup_background_map/"
+
+
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -297,7 +297,7 @@ class MerginPlugin:
 
     def current_project_sync(self):
         """Synchronise current Mergin project."""
-        self.manager.project_status(self.mergin_proj_dir, show_sync_button=True)
+        self.manager.project_status(self.mergin_proj_dir)
 
     def on_qgis_project_changed(self):
         """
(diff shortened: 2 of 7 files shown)


PeterPetrik (commented):
Looks good. 

for discussion:

- For situation when they are NO warnings, whatabout showing the green banner ":) You project is perfect :"-kind-of. OR/AND show sync button in green color?
- For situation when  they are Warniings, show sync button in "warning" color? 
- For me, I would show the same dialog (WITH sync button) all the time. People immediately forgot which button they clicked so they may be confused why there is no sync button.

saberraz:
Looks very good!

My suggestion is to have this dialogue displayed all the time when user clicks on Sync button. We can offer them an option to disable it and display it only through the Status button. But I have seen almost all users just press Sync and do not pay enough attention to the warnings/errors.

alexbruy:
> For situation when they are NO warnings, whatabout showing the green banner ":) You project is perfect :"-kind-of. OR/AND show sync button in green color?

While there can be no warnings from validations, some changes might trigger warning banner. For example, if user don't have writing rights for the project or there is an unfinished pull. I would vote for highlighting Sync button with the green color.

> My suggestion is to have this dialogue displayed all the time when user clicks on Sync button

It already works this way.

PeterPetrik (approved):

PR #373, opened 2022-04-09
Improve proxy handling and login UX

### LINKED ISSUE #297: Notify user about wrong credentials during login
When you type in incorrect credentials during login and click `ok`, login window disappears and user do not see any message about his credentials being incorrect. It is printed only in console.

I thought I am logged in actually.

alexbruy:
- fix parsing of excluded URL's (ref https://github.com/lutraconsulting/qgis-mergin-plugin/issues/309#issuecomment-978125062)
- test credentials and report if they are invalid (fix #297)
- try to use system proxy if QGIS uses "Default proxy" (ref https://github.com/lutraconsulting/qgis-mergin-plugin/issues/309#issuecomment-982423292)

### DIFF
--- a/Mergin/configuration_dialog.py
+++ b/Mergin/configuration_dialog.py
@@ -45,11 +45,24 @@ class ConfigurationDialog(QDialog):
         self.check_credentials()
 
     def accept(self):
+        # first check whether url is a corrent Mergin URL
         err_msg = validate_mergin_url(self.server_url())
         if err_msg:
-            msg = f"<font color=red> {err_msg} </font>"
+            msg = f"<font color=red>{err_msg}</font>"
             self.ui.test_status.setText(msg)
             return
+
+        # then try to login with the given credentials
+        username = self.ui.username.text()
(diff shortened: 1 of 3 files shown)


Mergin/configuration_dialog.py
```diff
@@ -45,11 +45,24 @@ def __init__(self):
...
+        # then try to login with the given credentials
+        username = self.ui.username.text()
+        password = self.ui.password.text()
+        proxy_config = get_qgis_proxy_config(self.server_url())
+        try:
+            mc = MerginClient(self.server_url(), None, username, password, get_plugin_version(), proxy_config)
+        except (URLError, LoginError, ClientError) as e:
+            msg = f"<font color=red>{e}</font>"
+            self.ui.test_status.setText(msg)
+            return
```
PeterPetrik, line 64: would be good to extract logic used here and in test_connection in the same routine so it is not duplicated, `  def test_connection(self):`

PeterPetrik (approved):

PR #383, opened 2022-04-21
add photo quality setting to mergin project properties

alexbruy:
Adds dropdown to Mergin project settings dialog with the following options: "Original", "High", "Medium" and "Low". By default "Original" setting is used to keep backward compatibility.

зображення

Refs lutraconsulting/input/issues/841.

### DIFF
--- a/Mergin/project_settings_widget.py
+++ b/Mergin/project_settings_widget.py
@@ -1,9 +1,10 @@
 import json
 import os
 from qgis.PyQt import uic
-from qgis.gui import QgsOptionsWidgetFactory, QgsOptionsPageWidget
 from qgis.PyQt.QtGui import QIcon
 from qgis.PyQt.QtWidgets import QFileDialog
+from qgis.core import QgsProject
+from qgis.gui import QgsOptionsWidgetFactory, QgsOptionsPageWidget
 from .utils import icon_path, mergin_project_local_path
 
 ui_file = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'ui', 'ui_project_config.ui')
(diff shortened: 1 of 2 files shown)


PeterPetrik (approved):

wonder-sk:
May I suggest that this new option will also get a dedicated group box, and a short explanatory label, e.g.: "This option is used by Input when adding pictures to the project (taken from camera or using existing picture from gallery)."

Mergin/project_settings_widget.py
```diff
@@ -28,6 +29,16 @@ class ProjectConfigWidget(ProjectConfigUiWidget, QgsOptionsPageWidget):
     def __init__(self, parent=None):
         QgsOptionsPageWidget.__init__(self, parent)
         self.setupUi(self)
+
+        self.cmb_photo_quality.addItem("Original", 0)
+        self.cmb_photo_quality.addItem("High", 1)
+        self.cmb_photo_quality.addItem("Medium", 2)
+        self.cmb_photo_quality.addItem("Low", 3)
```
wonder-sk, line 36: can we get some extra info on what the High/Medium/Low mean? either expected size in MB or expected size in pixels (or both)

alexbruy: Added information about approximate size, as I'm not sure about correct wording for pixel size. We have to take into account image orientation so it will be either width or height

PeterPetrik:
It wooudl be good to add expanatory text as Martin suggested

alexbruy:
Explanatory text added. Also added link to the selective sync documentation.

зображення

tomasMizera:
I suggest rewording the text for selective sync to "With selective sync enabled Input will not download photos from other surveyors to make synchronisation faster. Check out more in the *documentation*"

Otherwise, I believe we are ready to merge (afaik it passed testing by @jozef-budac ). @wonder-sk

wonder-sk:
Good suggestion @tomasMizera - let's update the text :-)

PR #387, opened 2022-04-28
MM Rebranding

tomasMizera:
PR replaces logos with new ones for the rebranding.

An additional icon is added at the beginning of the MM toolbar with the MM company icon - on click it opens the MM webpage.

I am wondering if we should not do the transition of toolbar icons in two phases so that users have a chance to get used to the new logo. I find the new toolbar visually way too different to do it in one go. Previous: 
image

Options for the new one:
 - a) image new logo + legacy icons
 - b) image new logo + legacy icons (but without legacy mergin logo) ~ we can make these icons smaller and have border just like on the previous option
 - c) image new logo + new icons from new icon pack

Which one would you prefer @wonder-sk / @PeterPetrik ?

---

There is also a new `QMessageBar` message informing about rebranding - this message is shown to authorised users. It is visible only once.

---

**Bonus question for @wonder-sk**: We also change the plugin's metadata - **name**, is that ok? Is it still going to be the same plugin in the QGIS plugin repository?

---

_Some screenshots:_

image

image

image

image


alexbruy:
> We also change the plugin's metadata - name, is that ok? Is it still going to be the same plugin in the QGIS plugin repository?

If I'm not wrong, if we change name it will be new plugin from the plugin repository point of view.

alexbruy:
@tomasMizera shouldn't we change dialog titles too? E.g. "Mergin Settings" to "MerginMaps Settings" and so on.

alexbruy (approved):
Code looks good to me.

Mergin/metadata.txt
```diff
@@ -1,6 +1,6 @@
 ; the next section is mandatory
 [general]
-name=Mergin
+name=Mergin Maps
```
PeterPetrik, line 3: keep the old name please

PeterPetrik (approved):

PeterPetrik:
very nice. 

I suggest to keep old name ("Mergin") for plugin name s it is same in the qgis plugin repo.

- for icons, I am easy with both a) -> c) (in phase 3) or immediately c).  I will leave it with you. I will not go with b) 
- also for notification, we can remove it in phase 3

tomasMizera:
Allright, let's go with the option a) in this phase.

PR #389, opened 2022-04-29
temporarily hide image quality settings from Mergin project properties

alexbruy:


tomasMizera (approved):

tomasMizera:
Note: we are hiding the element for the sake of rebranding - we will enable it when rebranding is released.

PR #391, opened 2022-05-02
Add logo with text to the plugin toolbar

alexbruy:
Add MerginMaps logo with text to the toolbar. Here is how it looks now
зображення


wonder-sk:
@alexbruy is it possible to make the logo a bit bigger so that it is easier to read? (maybe shave off some padding at the top & bottom of the SVG image that is used?)

alexbruy:
@wonder-sk yes it is possible and I showed both options to @tomasMizera. He suggested that we stay with version with padding for now.

I have updated SVG, without padding it looks like this
зображення

tomasMizera:
Yeah, but I am easy to also keep it like this :)

PR #399, opened 2022-05-09
fix child layer detection in value relation validation (fix #393)

### LINKED ISSUE #393: Since 2022.3 synching is throwing AttributeError in some projects
We have similar (cloned) projects with a so far working value-relation. In some projects we have now problems to synch since the last update (2022.3). This is probably due to different data in the projects.

I guess it is related to #380 

The Traceback looks like this: 
```
2022-05-03T16:07:33     WARNING    Traceback (most recent call last):
              File ".../AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\Mergin\plugin.py", line 346, in current_project_status
              self.manager.project_status(self.mergin_proj_dir)
              File ".../AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\Mergin\projects_manager.py", line 164, in project_status
              validation_results = validator.run_checks()
              File ".../AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\Mergin\validation.py", line 97, in run_checks
              self.check_value_relation()
              File ".../AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\Mergin\validation.py", line 262, in check_value_relation
              idx = child_layer.fields().indexFromName(str(cfg["Key"]))
             AttributeError: 'NoneType' object has no attribute 'fields'
```
In the project one relation is declared as Associatio between a point layer (fid) and non-geo-table (with the ID of point-layer). 
The association is used in the formula layout in an own tab.

alexbruy:
Relying on the layer source is not optimal, better to use layer id.

Fixes #393.

### DIFF
--- a/Mergin/validation.py
+++ b/Mergin/validation.py
@@ -41,6 +41,7 @@ class Warning(Enum):
     FIELD_IS_PRIMARY_KEY = 15
     VALUE_RELATION_LAYER_MISSED = 16
     INCORRECT_FIELD_NAME = 17
+    BROKEN_VALUE_RELATION_CONFIG = 18
 
 class MultipleLayersWarning:
     """Class for warning which is associated with multiple layers.
@@ -253,6 +254,10 @@ class MerginProjectValidator(object):
                 ws = layer.editorWidgetSetup(i)
                 if ws and ws.type() == "ValueRelation":
                     cfg = ws.config()
+                    if "Layer" not in cfg or "Key" not in cfg:
+                        self.issues.append(SingleLayerWarning(lid, Warning.BROKEN_VALUE_RELATION_CONFIG))
+                        continue
+
                     child_layer = next((v for k, v in self.layers.items() if k == cfg["Layer"]), None)
                     if child_layer is None:
                         self.issues.append(SingleLayerWarning(lid, Warning.VALUE_RELATION_LAYER_MISSED))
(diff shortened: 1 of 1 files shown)


Mergin/validation.py
```diff
@@ -253,9 +253,10 @@ def check_value_relation(self):
                 ws = layer.editorWidgetSetup(i)
                 if ws and ws.type() == "ValueRelation":
                     cfg = ws.config()
-                    child_layer = next((l for l in self.layers.values() if l.source() == cfg["LayerSource"]), None)
+                    child_layer = next((v for k, v in self.layers.items() if k == cfg["Layer"]), None)
```
PeterPetrik, line 256: are we sure cfg always has this key? or for safety add try..catch keyerror?

alexbruy: Yes, these keys should be there. Otherwise value relation will be not configured and will not work. Maybe it is good idea to check this too, though I'm not sure user will leave it in a such broken state.

Mergin/validation.py
```diff
@@ -253,9 +254,14 @@ def check_value_relation(self):
                 ws = layer.editorWidgetSetup(i)
                 if ws and ws.type() == "ValueRelation":
                     cfg = ws.config()
-                    child_layer = next((l for l in self.layers.values() if l.source() == cfg["LayerSource"]), None)
+                    if "Layer" not in cfg or "Key" not in cfg:
```
PeterPetrik, line 257: Where "Key" is used in the for loop below?

alexbruy: Right.

PeterPetrik (approved):

Issue #278, opened 2021-07-19
Allow sync of empty folder 


nicogodet:
Use case scenario : 
Create an empty folder "Photos" 
Create a layer with photo widget that store the picture using a relative path : `@project_home + '/Photos'`

If the folder Photos is empty, it will not be sync resulting in broken photo widget in Input

Work around : create an empty file inside Photos folder to sync it.

alexbruy:
Fixed in Input

geomenke:
Is there a plan to squash this bug?

tomasMizera:
Now it is more of an enhancement, than a bug. Mobile creates the folder if needed (as suggested before)

nicogodet:
I no longer use Input and Mergin but I would consider this fixed if Input creates the folder if it doesn't exists.
I think this was not the case when I opened this issue.

Feel free to close if you like

tomasMizera:
Hi @nicogodet, thanks for the reply, I am closing the ticket then. 
Out of curiosity, have you found a better alternative to Mergin Maps or aren't in need of surveying anymore? :)

PR #413, opened 2022-05-26
More validations for attachment widget configuration

alexbruy:
Check that data-defined default root uses `@project_home` and similar variables to create a path relative to project location.

Refs #278 and https://github.com/lutraconsulting/input/issues/1611.

### DIFF
--- a/Mergin/validation.py
+++ b/Mergin/validation.py
@@ -21,6 +21,7 @@ from .utils import (
 )
 
 INVALID_CHARS = re.compile("[\\\/\(\)\[\]\{\}\"\n\r]")
+PROJECT_VARS = re.compile("\@project_home|\@project_path|\@project_folder")
 
 
 class Warning(Enum):
@@ -42,6 +43,7 @@ class Warning(Enum):
     VALUE_RELATION_LAYER_MISSED = 16
     INCORRECT_FIELD_NAME = 17
     BROKEN_VALUE_RELATION_CONFIG = 18
+    ATTACHMENT_WRONG_EXPRESSION = 19
 
 class MultipleLayersWarning:
     """Class for warning which is associated with multiple layers.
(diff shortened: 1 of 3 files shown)


Mergin/validation.py
```diff
@@ -217,6 +219,14 @@ def check_attachment_widget(self):
                         if "UseLink" in cfg:
                             self.issues.append(SingleLayerWarning(lid, Warning.ATTACHMENT_HYPERLINK))
 
+                    # check that expression uses Mergin variables
+                    if "PropertyCollection" in cfg:
+                        if "properties" in cfg["PropertyCollection"]:
+                            if "propertyRootPath" in cfg["PropertyCollection"]["properties"] and "expression" in cfg["PropertyCollection"]["properties"]["propertyRootPath"]:
+                                formula = cfg["PropertyCollection"]["properties"]["propertyRootPath"]["expression"]
```
PeterPetrik, line 226: Maybe all this could be rewritten as 
```
try:
  formula = cfg["PropertyCollection"]["properties"]["propertyRootPath"]["expression"]
catch KeyError:
  return
```

Mergin/validation.py
```diff
@@ -221,11 +221,12 @@ def check_attachment_widget(self):
 
                     # check that expression uses Mergin variables
                     if "PropertyCollection" in cfg:
```
PeterPetrik, line 223: you can remove this if too

Mergin/validation.py
```diff
@@ -217,6 +219,15 @@ def check_attachment_widget(self):
                         if "UseLink" in cfg:
                             self.issues.append(SingleLayerWarning(lid, Warning.ATTACHMENT_HYPERLINK))
 
+                    # check that expression uses Mergin variables
+                    if "PropertyCollection" in cfg:
+                        try:
+                            formula = cfg["PropertyCollection"]["properties"]["propertyRootPath"]["expression"]
+                        except KeyError:
+                            continue
+                        if not PROJECT_VARS.search(formula):
```
PeterPetrik, line 228: sthis should be evaulated only if there is not cauth exception?

Mergin/validation.py
```diff
@@ -217,6 +219,14 @@ def check_attachment_widget(self):
                         if "UseLink" in cfg:
                             self.issues.append(SingleLayerWarning(lid, Warning.ATTACHMENT_HYPERLINK))
 
+                    # check that expression uses Mergin variables
+                    try:
+                        formula = cfg["PropertyCollection"]["properties"]["propertyRootPath"]["expression"]
+                    except KeyError:
+                        continue
+                    if not PROJECT_VARS.search(formula):
```
PeterPetrik, line 227: which KeyError, there is not formula defined?

PR #430, opened 2022-06-15
Snapping settings

alexbruy:
Add Input snapping settings to the QGIS project setting
зображення

Also add a validation for the case when "Follow QGIS snapping" mode is selected but snapping is not enabled in the project.

### DIFF
--- a/Mergin/project_settings_widget.py
+++ b/Mergin/project_settings_widget.py
@@ -35,9 +35,17 @@ class ProjectConfigWidget(ProjectConfigUiWidget, QgsOptionsPageWidget):
         self.cmb_photo_quality.addItem("Medium (approx. 1-2 Mb)", 2)
         self.cmb_photo_quality.addItem("Low (approx. 0.5 Mb)", 3)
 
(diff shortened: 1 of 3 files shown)


wonder-sk:
It would be good to include a link in the GUI to a page where this will be documented in the future :)

alexbruy:
> It would be good to include a link in the GUI to a page where this will be documented in the future :)

We have a place reserved for the link, but it is not clear where it will go. I can insert link using selective sync (https://merginmaps.com/docs/manage/selective_sync/) as an example, but not sure this is a right way to go.

PeterPetrik:
@alex-cit do you know the future link name? or it will be part of some existing page?

alex-cit:
I would reference this: https://merginmaps.com/docs/gis/features/#project-settings its overview of general project settings, including photo quality and selective sync.

alexbruy:
Linked to https://merginmaps.com/docs/gis/features/#project-settings.

varmar05 (approved):

Issue #432, opened 2022-06-20
When GeoPackage is in edit mode with pending changes and selecting Show local changes


saberraz:
...it would be good to ask user to either save or cancel changes. Currently, it does not show the pending changes made to the GeoPackage and not saved.

ldebek:
Can't reproduce this issue. It seems like it was fixed in the latest version of the plugin.
![save](https://user-images.githubusercontent.com/11883222/180780836-b5bbf4a5-9022-4efb-84fe-af49963fa5a2.PNG)

alexbruy:
Fixed with #447

PR #443, opened 2022-07-27
Added check for unsaved modifications in the project while showing lo…

ldebek:
Fix for the issue #432

### DIFF
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -355,9 +355,7 @@ class MerginPlugin:
 
         if unsaved_project_check(force_saving=True):
             iface.messageBar().pushMessage(
-                "Mergin",
-                "Project contains unsaved modifications, can not compute local changes",
-                Qgis.Warning
+                "Mergin", "Project contains unsaved modifications, can not compute local changes", Qgis.Warning
             )
             return
 
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -355,7 +355,10 @@ class MerginPlugin:
 
         if not unsaved_project_check(force_saving=True):
             iface.messageBar().pushMessage(
-                "Mergin", "Project contains unsaved modifications, can not compute local changes", Qgis.Warning
+                "Mergin",
+                "Project contains unsaved modifications, can not compute local changes. "
+                "Save unsaved edits and try again",
+                Qgis.Warning,
             )
             return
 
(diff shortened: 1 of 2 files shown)


Mergin/plugin.py
```diff
@@ -353,6 +353,12 @@ def view_local_changes(self):
             iface.messageBar().pushMessage("Mergin", "Current project is not a Mergin project.", Qgis.Warning)
             return
 
+        if not unsaved_project_check(force_saving=True):
+            iface.messageBar().pushMessage(
+                "Mergin", "Project contains unsaved modifications, can not compute local changes", Qgis.Warning
```
alexbruy, line 358: I would extend message a bit, e.g. "Project contains unsaved modifications, can not compute local changes. Save unsaved edits and try again".

alexbruy (approved):

Mergin/plugin.py
ldebek: Done

jozef-budac:
I am still able to see local change without saving feature using by hitting "View changes" in Project status dialog

https://user-images.githubusercontent.com/83582181/181713294-c2b7cd8d-e9a4-4868-98b7-631fed77b758.mp4

ldebek:
@jozef-budac @alexbruy Shall we then allow access to Project status only if user saved his pending changes?

jozef-budac:
Probably also @saberraz can ha have opinion about this

wonder-sk:
@ldebek I would prefer not to force users to save/revert changes when Sync button is clicked. How about this behavior:
- when user presses Sync, and some layers have unsaved changes, we would ask whether to save the pending changes with Yes/No/Cancel. If user clicks No, we would still open the project status dialog, but with a warning message bar on top, saying that some changes to layers are omitted because they are not saved yet (ideally also with a button to save the change)
- when user clicks show changes in popup menu, we would also ask about saving unsaved changes (Yes/No/Cancel), and also in this case if user clicks No, we would show local changes dialog, but with a message bar on top with a warning that some pending changes are omitted. (Instead of not allowing the dialog to be shown at all)

What do you think?

ldebek:
@wonder-sk What with a behaviour for contex menu action 'Show local changes'? Currently it shows warning with message "Project contains unsaved modifications, can not compute local changes. Save unsaved edits and try again". Shouldn't it work in the same way for a button in Project status window?

alexbruy:
@ldebek I think it should work in the same way as when opened via the button in Project Status: so show local changes and message bar with warning on top.

wonder-sk:
Yeah it should work the same way as the project status window - that's actually what I meant to describe in the second item in my comment, but I failed to describe it clearly :slightly_smiling_face:  - indeed with context menu action "show local changes" I would also first allow saving, and even if unsaved, show changes (with a warning message bar)

PR #447, opened 2022-07-29
Added warning in the local changes view if there are any pending chan…

ldebek:
…ges in project. (fix for the issue: #432)

### DIFF
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -286,8 +286,8 @@ class MerginPlugin:
 
     def create_new_project(self):
         """Open new Mergin Maps project creation dialog."""
-
-        if not unsaved_project_check():
+        proceed, replay_btn = unsaved_project_check()
+        if not proceed:
             return
         if not self.manager:
             QMessageBox.warning(None, "Create Mergin Maps Project", "Plugin not configured!")
@@ -353,14 +353,15 @@ class MerginPlugin:
             iface.messageBar().pushMessage("Mergin", "Current project is not a Mergin project.", Qgis.Warning)
             return
 
-        if not unsaved_project_check(force_saving=True):
+        proceed, reply_btn = unsaved_project_check()
+        if not proceed:
+            return
+        if reply_btn and reply_btn != QMessageBox.Yes:
             iface.messageBar().pushMessage(
                 "Mergin",
-                "Project contains unsaved modifications, can not compute local changes. "
-                "Save unsaved edits and try again",
+                "Project contains unsaved modifications, which won't be visible in the local changes view.",
                 Qgis.Warning,
             )
-            return
 
         mp = MerginProject(QgsProject.instance().homePath())
         push_changes = mp.get_push_changes()
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -39,6 +39,7 @@ from .sync_dialog import SyncDialog
 from .utils import (
     ClientError,
     LoginError,
+    check_mergin_subdirs,
     create_mergin_client,
     find_qgis_files,
     get_mergin_auth,
@@ -51,7 +52,7 @@ from .utils import (
     same_dir,
     unhandled_exception_message,
     unsaved_project_check,
-    check_mergin_subdirs,
+    UnsavedChangesStrategy,
 )
 
 from .mergin.merginproject import MerginProject
@@ -286,8 +287,8 @@ class MerginPlugin:
 
     def create_new_project(self):
         """Open new Mergin Maps project creation dialog."""
-        proceed, replay_btn = unsaved_project_check()
-        if not proceed:
+        check_result = unsaved_project_check()
+        if check_result == UnsavedChangesStrategy.HasUnsavedChanges:
             return
         if not self.manager:
             QMessageBox.warning(None, "Create Mergin Maps Project", "Plugin not configured!")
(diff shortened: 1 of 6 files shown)


Mergin/plugin.py
```diff
@@ -353,14 +353,15 @@ def view_local_changes(self):
             iface.messageBar().pushMessage("Mergin", "Current project is not a Mergin project.", Qgis.Warning)
             return
 
-        if not unsaved_project_check(force_saving=True):
+        proceed, reply_btn = unsaved_project_check()
+        if not proceed:
+            return
+        if reply_btn and reply_btn != QMessageBox.Yes:
```
alexbruy, line 359: I think better to show this message in the diff viewer dialog. In current implementation messagebar partialy hidden and not very readable

Mergin/project_status_dialog.py
```diff
@@ -204,7 +205,15 @@ def show_changes(self):
         if not self.changes_summary:
             self.ui.messageBar.pushMessage("Mergin", "No changes found in the project layers.", Qgis.Info)
             return
-
+        proceed, reply_btn = unsaved_project_check()
+        if not proceed:
+            return
+        if reply_btn and reply_btn != QMessageBox.Yes:
+            iface.messageBar().pushMessage(
```
alexbruy, line 212: Same as above. Better to show this message in the project satus dialog. In current implementation messagebar partialy hidden and not very readable

Mergin/utils.py
```diff
@@ -408,13 +408,12 @@ def get_new_qgis_project_filepath(project_name=None):
...
-def unsaved_project_check(force_saving=False):
+def unsaved_project_check():
     """
     Check if current QGIS project has some unsaved changes.
     Let the user decide if the changes are to be saved before continuing.
-    :return: True if previous method should continue, False otherwise.
-    If 'force_saving' flag is set to True then changes have to be saved before continuing.
-    :type: boolean
+    :return: True if previous method should continue, False otherwise and replay button (or None if no changes detected)
+    :type: boolean, QMessageBox.Yes|QMessageBox.No|QMessageBox.Cancel|None
```
alexbruy, line 416: Maybe we should rename Cancel to Ignore for better clarity? If I read code correctly Cancel does not cancel anything, it just tells to leave changes and pretend that project and layers are in the clean state.

alexbruy (changes requested):

Mergin/utils.py
```diff
@@ -408,13 +408,12 @@ def get_new_qgis_project_filepath(project_name=None):
...
 
-def unsaved_project_check(force_saving=False):
+def unsaved_project_check():
     """
     Check if current QGIS project has some unsaved changes.
     Let the user decide if the changes are to be saved before continuing.
-    :return: True if previous method should continue, False otherwise.
-    If 'force_saving' flag is set to True then changes have to be saved before continuing.
-    :type: boolean
+    :return: True if previous method should continue, False otherwise and replay button (or None if no changes detected)
```
alexbruy, line 415: Anothe way would be to return some enum value for each state: has unsaved changes, no unsaved changes, has changes but ignore.

Mergin/utils.py
ldebek: Actually "Cancel" button have a reject role, so it should result with not displaying Diff dialog.

Mergin/plugin.py
ldebek: Done.

Mergin/project_status_dialog.py
ldebek: Done, but message is showing in the diff viewer dialog (same as for comment above)

Mergin/utils.py
ldebek: Done.

Mergin/projects_manager.py
```diff
@@ -40,7 +41,8 @@ def unsaved_changes_check(project_dir):
         Check if current project is the same as actually operated Mergin project and has some unsaved changes.
         """
         if QgsProject.instance().fileName() in find_qgis_files(project_dir):
-            return True if unsaved_project_check() else False
+            check_result = unsaved_project_check()
+            return False if check_result == UnsavedChangesStrategy else True
```
alexbruy, line 45: Missed enum value?

Mergin/project_status_dialog.py
```diff
@@ -204,8 +203,13 @@ def show_changes(self):
         if not self.changes_summary:
             self.ui.messageBar.pushMessage("Mergin", "No changes found in the project layers.", Qgis.Info)
             return
-
-        self.close()
+        check_result = unsaved_project_check()
+        if check_result == UnsavedChangesStrategy.HasUnsavedChanges:
+            return
```
alexbruy, line 208: Not sure if returning is the right solution here. Let's imagine situation that some changes were made and saved, while some are not saved. Probably we still should show viewer with the warning in that case.

I'm not an UX expert, and would welcome second opinion here.

alexbruy:
Looks good to me. there is one typo to fix. Otherwise I'm happy with this.

Mergin/projects_manager.py
ldebek: Fixed.

jozef-budac:
I would go with with showing messagebox twice

Mergin/project_status_dialog.py
wonder-sk: As far as I understand from the code, this would open the "unsaved changes" dialog again, and only return prematurely if "Cancel" was clicked. This logic seems fine to me...

PR #448, opened 2022-08-09
allow reusing folders

### LINKED ISSUE #182: [Feature] Setup project connection
I'm not sure about the wording and maybe this can be done somehow.

Basically I have my account on a remove machine and locally.

The main issue is that I can only update the projects from the source that has uploaded the project. I cannot 'connect' my other instance to the project. I can only download and do some other option.

Could I simply 'link' a project to a folder. This way all my sessions could be able to synchronize the projects by simply setting up the connection to their shared sources.

Let me know if this is unclear. Thanks!

roya0045:
Simple proposal PR to allow reusing existing folders instead of having to delete and recreate them or create duplicates.

Open to suggestions for code changes.

Fixes #182


PeterPetrik:
@wonder-sk can you look at this PR please?

roya0045:
@wonder-sk did you have any reservation or wanted me to make changes to this? I think the risk is fairly low, though this was a quick pr that could be improved if needed.

wonder-sk:
hi @roya0045 - apologies for the long silence. I think this approach overall makes sense, but there are two important bits missing by the look at the code:
1. we should check first whether the project in the selected directory is the same as the project the user is trying to download
2. when reusing a project directory, we should also pull the latest changes from the server (instead of just leaving the directory with possibly outdated version of the project)

roya0045:
> hi @roya0045 - apologies for the long silence. I think this approach overall makes sense, but there are two important bits missing by the look at the code:
> 
> 1. we should check first whether the project in the selected directory is the same as the project the user is trying to download
> 2. when reusing a project directory, we should also pull the latest changes from the server (instead of just leaving the directory with possibly outdated version of the project)

I made a modified proposal, let me know what you think of it.

roya0045:
Tested on a more recent version and the following part is proving problematic https://github.com/MerginMaps/qgis-mergin-plugin/blob/93224742a112363f412354163f15d66752c0f251/Mergin/projects_manager.py#L191-L195

I have no clue why this exists.

PR #450, opened 2022-08-09
try except to handle missing ui elements

### LINKED ISSUE #440: Error after update 2022.5
Hi.
I had Mergin (qgis plug) version 2022.4.1 installed, both in QGIS 3.22 and 3.26, in windows 10. No problem.

But, after updating the mergin qgis plug to version 2022.5, I had errors when opening the project properties panel, where the GUI becomes unstable and the texts overlap, making it unusable.
In fact, the very update of the plug, to version 2022.5, ends with errors, but apparently then it lets you use the mergin and connect to the online account.

I tried a new, clean profile and no more problems, but I can't just use a new profile as the old one worked perfectly with all my settings.

If I go to the plug panel, and deactivate Mergin, the project properties panel works perfectly again.

image
image
image

https://user-images.githubusercontent.com/38449926/178317492-087c9dab-5e7e-4c6f-a76c-afb5415928e3.mp4

SO: windows 10
QGIS: 3.26 / 3.22 (osgeo)

roya0045:
In some instances it seems the modified ui files were not pushed in the last update.

This may be an edge case, but as a fix for those instances a simple try except prevent a QGIS crash or a corruption of the Property dialog.

Fixes #440

This bugfix might not be mandatory.

### DIFF
--- a/Mergin/project_settings_widget.py
+++ b/Mergin/project_settings_widget.py
@@ -38,15 +38,16 @@ class ProjectConfigWidget(ProjectConfigUiWidget, QgsOptionsPageWidget):
         quality, ok = QgsProject.instance().readNumEntry("Mergin", "PhotoQuality")
         idx = self.cmb_photo_quality.findData(quality) if ok else 0
         self.cmb_photo_quality.setCurrentIndex(idx if idx > 0 else 0)
+        try:
+            self.cmb_snapping_mode.addItem("No snapping", 0)
+            self.cmb_snapping_mode.addItem("Basic snapping", 1)
+            self.cmb_snapping_mode.addItem("Follow QGIS snapping", 2)
 
-        self.cmb_snapping_mode.addItem("No snapping", 0)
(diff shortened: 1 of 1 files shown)


PeterPetrik (commented):
I do not think that this is the way to go, since actually we would need to try-catch all lines :) and also if the UI is not updated, we would better like user to reinstall the plugin/restart QGIS and fix it. Otherwise the other error can come from any other place.

I suggest to close this PR

roya0045:
> I do not think that this is the way to go, since actually we would need to try-catch all lines :) and also if the UI is not updated, we would better like user to reinstall the plugin/restart QGIS and fix it. Otherwise the other error can come from any other place.
> 
> I suggest to close this PR

I agree, I don't know why reinstalling the plugin in some instances still doesn't update ui elements in my case. I just wanted to propose the PR in the even where this may be a more common error as it seems I wasn't the only one with that.

PR #453, opened 2022-08-23
Make icons more visible with dark themes (fix #434)

### LINKED ISSUE #434: Dark mode
In case user have dark mode on PC, QGIS plugin icons are hard to see:

plugin

It would be good to have white icon in case QGIS uses dark mode.

alexbruy:
Add iconset for dark modes.

image

Fixes #434.


varmar05:
@alexbruy is it possible to change link color (sign up)

alexbruy:
It is possible, but all colors usually are defined at the app theme level (in this case in the QGIS themes) and changing it will be inconsistent with the rest of the app.

varmar05 (approved):

jozef-budac:
I have one question about this. When my OS have dark theme and in QGIS there is 'defaut' as a UI theme, QGIS uses dark theme same as OS but our plugin uses light theme. 
If I set dark theme (Night mapping) in QGIS, our plugin uses dark theme also. 
Is this correct?

alexbruy:
> I have one question about this. When my OS have dark theme and in QGIS there is 'defaut' as a UI theme, QGIS uses dark theme same as OS but our plugin uses light theme.

I'm afraid we can't detect OS-level theme, only QGIS setting is taken into account.

jozef-budac:
@alexbruy thanks!

tomasMizera (commented):
The icon pack works well with both `Night Mapping` and `Blend of Gray` themes.

The issue that @jozef-budac found is that when OS has dark mode enabled, QGIS becomes dark as well, but the `UI/UITheme` stays `default`. 
It stays default, because `QGIS OS dark mode` != `Night Mapping`. It is a different theme. You can see it on the example below:

<table>
  <tr>
    <th>Night Mapping</th>
    <th>Dark mode from OS</th>
  </tr>
  <tr>
    <td>image</td>
    <td>image</td>
  </tr>
 </table>

> Checkboxes and buttons are different

**As a result, our plugin icons stay black in OS dark mode**.
We would need to detect if underlying OS has dark mode enabled. It seems that QGIS API  _contains_ a way how to determine that, see `qgsqative.h`. One would call `QgsGui::instance().nativePlatformInterface().hasDarkTheme()` 
Each system (win/linux/mac) has its implementation of `qgs[win/linux/mac]native.h` to detect the dark mode.  


Sad thing is that `nativePlatformInterface` is not exposed to python API (I guess from security concerns?) .. One way to go would be to expose the `QgsGui::instance().nativePlatformInterface().hasDarkTheme()` value somewhere where python API has access (QgsSettings?). 

That being said, I have not managed to find where the `hasDarkTheme` function is used in QGIS code. 
@wonder-sk do you have any idea how to tackle this one? Shall we merge what we have now and create new ticket for the OS dark mode? 

See code here: https://github.com/qgis/QGIS/pull/8309/commits/b4fd0275e124fa8c644e73d6ebe4472fab79999f#diff-a6c6864f28b601d260d5aff00f149039a0df622797f72f15311d052b1a38b77fR2005-R2013

See discussion here: https://github.com/qgis/QGIS/pull/8309 
See ticket: https://github.com/qgis/QGIS/issues/33079

alexbruy:
Note that `hasDarkMode` currently implemented only for Mac. Also this API is not exposed to Python.

wonder-sk:
Hmm it is indeed a bit tricky :-)
As far as I understand, `hasDarkMode()` was used in some older releases (to work around bad Qt handling) but it's not used anymore. When you have "default" mode, it can be either light or dark, depending on the OS palette - and then one can explicitly switch to a different theme (night mapping / blend of gray).

I think I would use some kind of heuristics (using inspiration from https://github.com/qgis/QGIS/commit/8354915f0b7571c0fefa383b5e9c884e202d5d5e):

```python
pal=QgsApplication.instance().palette()
c_background = pal.color(QPalette.Window)
c_foreground = pal.color(QPalette.WindowText)
```
and use `c_background` or `c_foreground` to decide whether light or dark theme is more appropriate...

tomasMizera:
That sounds good. I tested it with logic here: https://stackoverflow.com/a/51567564/7875594 and it seems to work fine for me. I believe we only need `c_background` to make the decision.

I would still prefer to first check for `Night Mapping` || `..Gray` theme and only if the theme is `default` check for colours as Martin suggested.

alexbruy:
@jozef-budac may I ask you to retest this branch with OS dark theme?

Mergin/utils.py
```diff
@@ -1169,3 +1168,21 @@ def test_server_connection(url, username, password):
...
     return result
+
+
+def is_dark_theme():
+    """
+    Checks whether dark theme is used:
+    - first check theme used by QGIS and if it is "default" then
+    - try to detect if OS-level theme is dark
+    """
+    settings = QSettings()
```
tomasMizera, line 1179: Shouldn't this be `QgsSettings`?

jozef-budac:
@alexbruy  it looks ok now. But I tested version "add heurictic to detect OS-level dark mode". The last commit failed to build package

alexbruy:
@jozef-budac thanks! Last commit is a minor typo fix, which does not affect functionality.

Issue #184, opened 2020-12-04
Package the project grid shift with the project


saberraz:
It will be a good to have an option during upload, for the plugin to see if the project is using a grid shift. If so, prompt the user (maybe with some text saying: if not sure, press YES), if they want to use the grid shift in their project.

It then copies the grid shift for the current project from QGIS proj folder to the project folder and uploads. See: https://help.inputapp.io/howto/proj

saberraz:
See https://github.com/lutraconsulting/qgis-mergin-plugin/issues/197

alexbruy:
Fixed in #455

PR #455, opened 2022-08-29
Datum shift grids support

### LINKED ISSUE #444: Check that all required grid shift files are in the project
Some Mergin projects may need to include datum shift grid files to correctly do CRS transformations (https://dev.merginmaps.com/docs/gis/proj/).

We should try to check if any of the used transformations reference grid files, and those that do should have the grid file included in `proj` sub-folder. If any of the grid files are missing, it should be reported as a validation issue.

Related packaging ticket: #184

alexbruy:
Adds support for datum shift grids:

- when packaging or using current project as is plugin will try to copy all requred datum shift grids in the MerginMaps project directory
- added new validation to check that all requred datum shift grids are in the "proj" subdirectory of the MerginMaps project directory. If there are missed grids it is possible to try to fix this issue by clicking on the "Fix" link.

Fix #444, #184


Mergin/validation.py
```diff
@@ -310,6 +320,12 @@ def check_snapping(self):
                 # snapping in Input using QGIS setting is disabled but project has snapping activated
                 self.issues.append(MultipleLayersWarning(Warning.MERGIN_SNAPPING_NOT_ENABLED))
 
+    def check_datum_shift_grids(self):
+        grids = get_datum_shift_grids()
+        for grid in grids.keys():
+            if not os.path.exists(os.path.join(self.qgis_proj_dir, "proj", grid)):
```
varmar05, line 326: I found "proj" in two cases, maybe worth defining util function to get this directory for easier readability?

Mergin/validation.py
```diff
@@ -310,6 +320,12 @@ def check_snapping(self):
...
+                self.issues.append(MultipleLayersWarning(Warning.MISSING_DATUM_SHIFT_GRID))
```
varmar05, line 327: do we need multiple warnings if there are more grids missing?

Mergin/validation.py
```diff
@@ -356,3 +372,5 @@ def warning_display_string(warning_id):
         return "Snapping is currently disabled in this QGIS project, it will be thus disabled in Mergin Maps Input"
     elif warning_id == Warning.MERGIN_SNAPPING_NOT_ENABLED:
         return "Snapping is currently enabled in this QGIS project, but not enabled in Mergin Maps Input"
+    elif warning_id == Warning.MISSING_DATUM_SHIFT_GRID:
+        return "Required datum shift grid is missed. <a href='#fix_datum_shift_grids'>Fix.</a>"
```
varmar05, line 376: "... is missing. Fix issue."

Mergin/repair.py
```diff
@@ -0,0 +1,16 @@
+from qgis.core import QgsProject
+
+from .utils import copy_datum_shift_grids
+
+
+def fix_datum_shift_grids(mp):
+    qgis_proj_dir = QgsProject.instance().absolutePath()
+    if mp is not None:
+        qgis_proj_dir = mp.dir
```
varmar05, line 9: what if qgis project file is not in root of mergin project?

Mergin/repair.py
```diff
@@ -0,0 +1,16 @@
...
+
+    missed_files = copy_datum_shift_grids(qgis_proj_dir)
+
+    if missed_files:
+        return False, f"Following grids were not found in the QGIS folder: {','.join(missed_files)}"
+
+    return True, None
```
varmar05, line 16: I suggest to add docstring with description what we are returning. Also error message might be enough information to return?

Mergin/project_status_dialog.py
```diff
@@ -68,6 +71,11 @@ def __init__(
         self.treeStatus.expandAll()
         self.changes_summary = push_changes_summary
 
+        try:
+            self.module = importlib.import_module("Mergin.repair")
+        except ImportError:
+            self.module = None
```
varmar05, line 77: when this can happen?

varmar05 (changes requested):

Mergin/validation.py
alexbruy: I don't think we need many identical warnings in this case. We do the same in some other cases, show only one waring for similar issues.

Mergin/project_status_dialog.py
alexbruy: In theory it should not happen. I can remove it if you think it is too much.

Mergin/validation.py
wonder-sk: "Fix the issue"

And if there's enough space for more explanation, I would add another sentence, e.g. "Reprojection may not work correctly."

Mergin/validation.py
```diff
@@ -310,6 +320,12 @@ def check_snapping(self):
                 # snapping in Input using QGIS setting is disabled but project has snapping activated
                 self.issues.append(MultipleLayersWarning(Warning.MERGIN_SNAPPING_NOT_ENABLED))
 
+    def check_datum_shift_grids(self):
+        grids = get_datum_shift_grids()
+        for grid in grids.keys():
+            if not os.path.exists(os.path.join(self.qgis_proj_dir, "proj", grid)):
+                self.issues.append(MultipleLayersWarning(Warning.MISSING_DATUM_SHIFT_GRID))
+
```
wonder-sk, line 328: Is it possible to also show the list of missing grid files?

Mergin/test/test_utils.py
```diff
@@ -55,6 +62,23 @@ def test_table_schema_changed(self):
...
+        ops = QgsDatumTransform.operations(crs_a, crs_b)
+        self.assertTrue(len(ops) > 0)
+        proj_str = ops[0].proj
+
+        context = QgsCoordinateTransformContext()
+        context.addCoordinateOperation(crs_a, crs_b, proj_str)
+        QgsProject.instance().setTransformContext(context)
+
+        grids = get_datum_shift_grids()
+        self.assertEqual(len(grids), 1)
```
wonder-sk, line 80: can we also check for the file name of the grid?

Mergin/repair.py
```diff
@@ -0,0 +1,16 @@
+from qgis.core import QgsProject
+
+from .utils import copy_datum_shift_grids
+
+
+def fix_datum_shift_grids(mp):
```
wonder-sk, line 6: Can we have a docstring :)

Mergin/project_status_dialog.py
wonder-sk: +1 to do ordinary import at the top of the file to avoid confusion

Mergin/project_status_dialog.py
```diff
@@ -213,3 +221,28 @@ def show_changes(self):
...
         dlg_diff_viewer.exec_()
+
+    def link_clicked(self, url):
+        if self.module is None:
+            self.ui.messageBar.pushMessage("Mergin", f"Repair module not found", Qgis.Warning)
+            return
+
+        function_name = url.toString().strip("#")
+        if hasattr(self.module, function_name):
+            ok, msg = getattr(self.module, function_name)(self.mp)
```
wonder-sk, line 232: I would prefer to make this more explicit - e.g. if the link was "#fix_datum_shift_grids" then we will directly call `fix_datum_shift_grids()` - otherwise it's difficult to know what are the supported fixes (and by looking for references to the function in IDE it may look like the `fix_datum_shift_grids()` is unused...

Mergin/project_status_dialog.py
```diff
@@ -37,54 +40,54 @@ def __init__(
...
+            for msg in info_text:
+                lbl = QLabel(msg)
+                lbl.setSizePolicy(QSizePolicy.Expanding, QSizePolicy.Preferred)
+                lbl.setWordWrap(True)
+                self.ui.messageBar.pushWidget(lbl, Qgis.Warning)
+
+            self.validate_project()
+
+            try:
+                self.module = importlib.import_module("Mergin.repair")
```
varmar05, line 88: I am still not sure why we cannot have ordinary import on top of file?

Mergin/repair.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    Copies datum shift grids to the MerginMaps project directory.
+    Returns None on success and error message if some grids were not
+    copied.
+    """
+    if mp is None:
+        return "Invalid MerginMaps project"
```
wonder-sk, line 13: I think the right way of spelling MM is two separate words:
```suggestion
        return "Invalid Mergin Maps project"
```

Mergin/create_project_wizard.py
```diff
@@ -493,9 +494,16 @@ def accept(self):
...
 
+            # copy datum shift grids
+            package_datum_grids(os.path.join(self.project_dir, "proj"))
+
         elif self.init_page.cur_proj_no_pack_btn.isChecked():
             cur_proj = QgsProject.instance()
             cur_proj.write()
+
+            # copy datum shift grids
+            package_datum_grids(os.path.join(QgsProject.instance().homePath(), "proj"))
```
wonder-sk, line 505: why not like this?
```suggestion
            package_datum_grids(os.path.join(self.project_dir, "proj"))
```
The problem is that `homePath()` can be arbitrary user-defined directory...

wonder-sk (approved):
:+1:  Thanks a lot!

PR #466, opened 2022-10-26
Workspaces

uclaros:
Requires https://github.com/MerginMaps/mergin-py-client/pull/150

Changes in classes `class MerginGroupItem(QgsDataCollectionItem):` and `class MerginRootItem(QgsDataCollectionItem):` are easier to review side-by-side before and after, the diff is messy.

If server is old and does not support workspaces, the groups in browser are reduced to two: My projects and Shared with me:
app dev

If server is CE, there are no groups and the global namespace appears in brackets:
ce dev

If server is EE, the active workspace is displayed in brackets and an option to switch workspace is offered:
Peek 2022-10-26 13-07

Switching between old/ce/ee servers should be transparent to the user:
Peek 2022-10-26 13-09

When there are no workspaces the user gets a warning with a link to mergin:
image

~Realized while taking screenshots:~ Fixed
- [x] ~*Apparently, when switching back to EE server, the last used workspace is not preserved, will investigate.*~
- [x] ~*Mergin Maps [None]* looks ugly, should probably remove brackets when there is no workspace~
- [x] ~*Create new project* should not be available when there is no workspace~


PeterPetrik:
👏 can you please add some screenshots/gifs?

uclaros:
> clap can you please add some screenshots/gifs?

Done (and found a couple of bugs while doing so)

Mergin/clone_project_dialog.py
```diff
@@ -1,29 +1,66 @@
...
-        self.ui.buttonBox.button(QDialogButtonBox.Ok).setEnabled(bool(self.ui.edit_project_name.text()))
+        enabled = bool(self.ui.edit_project_name.text()) and not self.ui.warningMessageLabel.isVisible()
+        self.ui.buttonBox.button(QDialogButtonBox.Ok).setEnabled(enabled)
+
+    def workspace_changed(self):
+        is_writable = bool(self.ui.projectNamespace.currentData(Qt.UserRole))
+        if is_writable:
+            msg = ""
+        else:
+            msg = "You do not have write permissions for this workspace"
```
varmar05, line 58: Not quite sure what means "writable". If user is a workspace writer, then he has permissions to push to all projects. If not, he can still write to projects which he was granted write permissions for. 

For creating new projects in workspace you need to be workspace admin. (However it used to be organization writer)

Mergin/create_project_wizard.py
```diff
@@ -112,11 +112,27 @@ def initializePage(self):
...
 
     def populate_namespace_cbo(self):
-        self.project_owner_cbo.addItem(self.parent.username)
-        if self.parent.user_organisations:
-            self.project_owner_cbo.addItems(
-                [o for o in self.parent.user_organisations if self.parent.user_organisations[o] in ["admin", "owner"]]
-            )
+        if self.parent.workspaces is not None:
+            for ws in self.parent.workspaces:
+                try:
```
varmar05, line 117: these check are in different file as well, is it possible to have some check function instead?

Mergin/plugin.py
```diff
@@ -262,6 +265,46 @@ def configure(self):
...
             self.show_browser_panel()
 
+    def show_no_workspaces_dialog(self):
+        msg = (
+            "Workspace is a place to store your projects and share them with your colleagues. "
+            "Click on the button below to create one. \n\n"
+            "A minimum of one workspace is required to use Mergin Maps."
+        )
+        msg_box = QMessageBox(QMessageBox.Critical, "You do not have any workspace", msg, QMessageBox.Close)
+        create_button = msg_box.addButton("Create workspace", msg_box.ActionRole)
```
varmar05, line 275: There can be EE deployment where creation of new workspaces is prohibited (https://ee.dev.merginmaps.com/config).
If that is a case only mergin super user can create new workspaces.

Mergin/plugin.py
```diff
@@ -262,6 +265,46 @@ def configure(self):
...
+        create_button.clicked.disconnect()
+        create_button.clicked.connect(self.open_configured_url)
+        msg_box.exec_()
+
+    def set_current_workspace(self, workspace):
+        settings = QSettings()
+        self.current_workspace = workspace
+        settings.setValue("Mergin/lastUsedWorkspace", str(workspace))
```
varmar05, line 283: is not is safer to use workspace_id? Just in case we allow renaming in future. Or workspace is removed and recreated again.

Mergin/plugin.py
```diff
@@ -295,8 +338,22 @@ def create_new_project(self):
...
 
         user_info = self.mc.user_info()
+        try:
+            workspaces = self.mc.workspaces_list()
+            if not workspaces:
+                self.show_no_workspaces_dialog()
+                self.current_workspace = None
+                return
+        except (URLError, ClientError) as e:
+            ns = self.mc.global_namespace()
```
varmar05, line 348: this only exists for CE server, no?

Mergin/plugin.py
```diff
@@ -461,7 +542,19 @@ def open_project(self):
 
     def clone_remote_project(self):
         user_info = self.mc.user_info()
-        dlg = CloneProjectDialog(username=user_info["username"], user_organisations=user_info.get("organisations", []))
+        try:
```
varmar05, line 545: I noticed this before, maybe refactor?

Mergin/plugin.py
```diff
@@ -461,7 +542,19 @@ def open_project(self):
...
+        except (URLError, ClientError) as e:
+            ns = self.mc.global_namespace()
+            workspaces = []  # todo: handle bad server response?
+            if ns:
+                workspaces.append(ns)
+            else:
+                workspaces = None
+
+        dlg = CloneProjectDialog(
+            user_info=user_info, workspaces=workspaces, default_workspace=self.project["namespace"]
```
varmar05, line 556: is this source or destination workspace?

Mergin/plugin.py
uclaros: Yes, that's a good point

Mergin/plugin.py
uclaros: Yes, and should be removed with the upcoming API, since the global namespace will be returned by user info and workspaces endpoint. Then we will not distinguish between ce and ee

Mergin/plugin.py
uclaros: `workspaces` is the list of available workspaces, `default_workspace` is the workspace that we want to be pre-selected in the workspace combobox

Mergin/clone_project_dialog.py
uclaros: OK, I got this wrong I guess. So let's get this straight.
In order to be able to clone into a workspace/namespace:
- in CE/EE the user id needs to be in *admins* or *owners*
- in old, the user needs to be in organization's *writers*?
The existing code prior to this PR was checking against organization's *admin* and *owner*
Can you confirm?

Mergin/plugin.py
uclaros: Is this something we should handle in the plugin though? What should we do in that case? Skip the *create workspace* button and append *Contact your administrator* to the message?

Mergin/clone_project_dialog.py
varmar05: - CE/EE - correct
- old: for destination namespace either it was user's namespace or user was organisation writer (checked latest server master)

Mergin/plugin.py
varmar05: Not sure, server will return 403. I can add some error message there. So if you just display that message, might be good to keep it.

Mergin/plugin.py
```diff
@@ -609,7 +706,20 @@ def submit_logs(self):
 
     def clone_remote_project(self):
         user_info = self.mc.user_info()
-        dlg = CloneProjectDialog(username=user_info["username"], user_organisations=user_info.get("organisations", []))
+        try:
```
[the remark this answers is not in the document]
uclaros: `class MerginRemoteProjectItem` and `class MerginLocalProjectItem` have too much in common apart from this method and could be refactored to derive from a single class, but I'd rather have this done in a separate PR

Mergin/plugin.py
uclaros: ok, I'll change the url to `/workspaces` (now it's pointing to `/) then. If the user sees a 403 he will contact his admin anyway! :)

tomasMizera (commented):
Great job @uclaros !!! 

Some notes about the switch workspace dialog:
 - will there be search if number of workspaces is higher than X? (5?) 
 - that dialog has title "Dialog", let's name it "Switch workspace"
 - can you replace the MM logo from this dialog with the one we have in the `configure` dialog (with text)? 

Screenshot 2022-11-02 at 09 47 54

uclaros:
image

Mergin/plugin.py
uclaros: fixed

Mergin/plugin.py
uclaros: changed url to `/workspaces`

tomasMizera:
@varmar05 you have unresolved review comments here. Can you check them, please?

varmar05 (approved):

PR #469, opened 2022-11-02
find project / explore public projects

uclaros:
Added _Find project_ and _Explore public projects_ dialogs, accessible via the browser's context menus.

For EE/CE:
image

For old servers:
image

Find projects looks like this:
image

The switch workspaces text is only visible is there are more workspaces available.

Explore public projects looks like this:
image


uclaros:
I updated the dialogs to use paginated results and fetch more when the srollbar reaches the bottom.
Requests are still on main thread so the UI is blocked while fetching.
To mitigate this I used a hybrid approach that first filters on server side, then if all results are fetched, subsequent searches starting with the same string are done locally!

uclaros:
Results are fetched in a background thread now.
Local filtering when all results are fetched still applies.

tomasMizera:
Let's merge this to have only one branch :)

PR #474, opened 2022-12-05
project name validation

alexbruy:
when creating a new project check that project name does not contain invalid characters


Mergin/test/test_utils.py
```diff
@@ -90,6 +90,81 @@ def test_datum_shift_grids(self):
...
+            ("projAUXect", True),
+            ("CONproject", True),
+            ("projectCON", True),
+            ("support", False),
+            ("helpdesk", False),
+            ("input", False),
+            ("lutraconsulting", False),
+            ("lutra", False),
+            ("merginmaps", False),
+            ("mergin", False),
```
varmar05, line 162: what about `admin` like? and maybe forbid `sales`?

alexbruy: Added both

Mergin/test/test_utils.py
```diff
@@ -90,6 +90,83 @@ def test_datum_shift_grids(self):
...
+            ("admin", False),
+            ("sales", False),
```
tomasMizera, line 164: Can you please add the same Chinese/Hebrew project names as you did for mobile https://github.com/MerginMaps/input/pull/2449/files#diff-33c6f9baefe15bf8b0b47c49256319fad26445333a6bf988a5f797ef60dae96eR232-R233 ? thanks!

tomasMizera (approved):
Code is looking good! :)

varmar05:
@tomasMizera I am wondering if we should not move this check to py-client as it would be useful for CLI as well
https://github.com/MerginMaps/mergin-py-client/blob/master/mergin/cli.py#L193

PR #475, opened 2022-12-08
Use new api/logic for _attention needed_

uclaros:
Requires https://github.com/MerginMaps/mergin-py-client/pull/152
Not yet tested


alexbruy (approved):
Code looks good to me

PR #489, opened 2023-03-31
Tracking setup for mobile application

alexbruy:
Add a new group to the Mergin project properties widget to setup tracking. User can activate/deactivate tracking and select desired tracking precision.
image
Tracking configuration then saved in the QGIS project file and will be used by mobile application.

If this is a first activation of tracking a new GPKG file called `tracking_layer.gpkg` with a `LineStringZM` layer will be created in the Mergin project folder and added to the project. Note, we require project to be saved before enabling tracking as otherwise we can not determine where to put GPKG file.

Peek 2023-03-31 12-45

### DIFF
--- a/Mergin/project_settings_widget.py
+++ b/Mergin/project_settings_widget.py
@@ -11,11 +11,18 @@ from qgis.core import (
     QgsFeature,
     QgsFeatureRequest,
     QgsExpression,
-    QgsVectorLayer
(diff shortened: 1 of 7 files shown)


alexbruy:
Suggestion applied (they were not presented in the specification).

As for the issue, I can't reproduce it on Linux and Windows
Peek 2023-04-27 09-02

PR #494, opened 2023-05-23
add download vector tiles algorithm

alexbruy:
Add algorithm to download vector tiles from the remote server. This is a port of the native QGIS alorithm for older QGIS versions.
image
Also add "Export to MBTiles" entry to the vector tile layer context menu to make it more accessible
image

### DIFF
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -478,8 +478,11 @@ class MerginPlugin:
             self.enable_toolbar_actions()
 
     def add_context_menu_actions(self, layers):
+        provider_names = ("vectortile")
+        if Qgis.versionInt() >= 33200:
+            provider_names = ("xyzvectortiles", "arcgisvectortileservice", "vtpkvectortiles")
         for l in layers:
-            if l.dataProvider().name() == "xyzvectortiles":
+            if l.dataProvider().name() in provider_names:
                 self.iface.addCustomActionForLayer(self.action_export_mbtiles, l)
 
     def unload(self):
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -167,7 +167,7 @@ class MerginPlugin:
             self.action_show_changes.triggered.connect(self.view_local_changes)
 
             # add layer context menu action for downloading vector tiles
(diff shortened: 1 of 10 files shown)


wonder-sk (approved):

tomasMizera:
How about renaming the item title from "export to MBTiles" to "make available offline" or something similar @wonder-sk ?

wonder-sk:
> How about renaming the item title from "export to MBTiles" to "make available offline" or something similar @wonder-sk ?

"Make available offline..." is a good suggestion - less technical, easier to understand!

tomasMizera:
@alexbruy can you adjust the title in the context menu please? `Export to MB Tiles` -> `Make available offline...`

Mergin/plugin.py
```diff
@@ -165,6 +166,12 @@ def initGui(self):
             self.iface.addCustomActionForLayerType(self.action_show_changes, "", QgsMapLayer.VectorLayer, True)
             self.action_show_changes.triggered.connect(self.view_local_changes)
 
+            # add layer context menu action for downloading vector tiles
+            self.action_export_mbtiles = QAction("Make available offline…", self.iface.mainWindow())
```
tomasMizera, line 170: @alexbruy can we have this action only for vector tiles layers? Other layers are not supported anyways

alex-cit:
Using the artifact from here: https://github.com/MerginMaps/qgis-mergin-plugin/actions/runs/5711118032 , I am getting this error when using the *Make available offline* function on the default mergin maps vector tile basemap.
I tried it in QGIS 3.22 and QGIS 3.28.9 using new profiles with the same result.

`Traceback (most recent call last):
File "C:\Users/sarah/AppData/Roaming/QGIS/QGIS3\profiles\test-qgis328/python/plugins\Mergin\processing\algs\download_vector_tiles.py", line 137, in prepareAlgorithm
if layer.type() != Qgis.QgsMapLayerType.VectorTileLayer:
AttributeError: type object 'Qgis' has no attribute 'QgsMapLayerType'`

alexbruy:
@alex-cit, looks like you are using old aritfact, the latest on is https://github.com/MerginMaps/qgis-mergin-plugin/suites/14685256926/artifacts/833590555

PR #536, opened 2023-10-25
[WIP] Switch project sources from GPKG to Postgre based on DBSync config

JanCaha:
Fixes MerginMaps/mergin-db-sync#11

Needs an icon!

### DIFF
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -40,6 +40,7 @@ from .diff_dialog import DiffViewerDialog
 from .project_settings_widget import MerginProjectConfigFactory
 from .projects_manager import MerginProjectsManager
 from .sync_dialog import SyncDialog
+from .switch_sources_dialog import ProjectUsePostgreConfigWizard
 from .configure_sync_wizard import DbSyncConfigWizard
 from .remove_project_dialog import RemoveProjectDialog
 from .utils import (
@@ -155,7 +156,13 @@ class MerginPlugin:
                 add_to_menu=True,
                 add_to_toolbar=None,
             )
-
+            self.action_switch_sources = self.add_action(
+                "database-cog.svg",
+                text="Switch DBSync Sources",
+                callback=self.switch_sources,
+                add_to_menu=True,
+                add_to_toolbar=None,
+            )
             self.enable_toolbar_actions()
 
         self.post_login()
@@ -323,6 +330,38 @@ class MerginPlugin:
         wizard = DbSyncConfigWizard(project_name)
         if not wizard.exec_():
             return
+
+    def switch_sources(self):
+        project_path = QgsProject.instance().homePath()
+        if not project_path:
+            iface.messageBar().pushMessage("Mergin", "Project is not saved, please save project first", Qgis.Warning)
+            return
+
+        if not check_mergin_subdirs(project_path):
+            iface.messageBar().pushMessage(
+                "Mergin", "Current project is not a Mergin project. Please open a Mergin project first.", Qgis.Warning
+            )
+            return
+
+        mp = MerginProject(project_path)
+        try:
+            project_name = mp.metadata["name"]
+        except InvalidProject as e:
+            iface.messageBar().pushMessage(
+                "Mergin", "Current project is not a Mergin project. Please open a Mergin project first.", Qgis.Warning
+            )
+            return
+
+        dialog = ProjectUsePostgreConfigWizard(self.iface)
+        if not dialog.exec():
+            return
+
+        if check_mergin_subdirs(dialog.new_project_parent_folder()):
+            iface.messageBar().pushMessage(
+                "Mergin",
+                "The updated project should not be saved within Mergin directory. Please move it elsewhere.",
+                Qgis.Warning,
+            )
 
     def show_no_workspaces_dialog(self):
         msg = (
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -360,13 +360,6 @@ class MerginPlugin:
         if not dialog.exec():
             return
 
-        if check_mergin_subdirs(dialog.new_project_parent_folder()):
-            iface.messageBar().pushMessage(
-                "Mergin",
-                "The updated project should not be saved within Mergin directory. Please move it elsewhere.",
-                Qgis.Warning,
-            )
-
     def show_no_workspaces_dialog(self):
         msg = (
             "Workspace is a place to store your projects and share them with your colleagues. "
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -47,6 +47,7 @@ from .utils import (
     ServerType,
     ClientError,
     LoginError,
+    InvalidProject,
     check_mergin_subdirs,
     create_mergin_client,
     find_qgis_files,
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -352,6 +352,10 @@ class MerginPlugin:
             )
             return
 
+        check_result = unsaved_project_check()
+        if check_result == UnsavedChangesStrategy.HasUnsavedChanges:
+            return
+
         dialog = ProjectUsePostgreConfigWizard(self.iface)
         if not dialog.exec():
             return
--- a/Mergin/switch_sources_dialog.py
+++ b/Mergin/switch_sources_dialog.py
@@ -1,0 +1,202 @@
+import os
+import yaml
+from dataclasses import dataclass, field
+import typing
+import re
+
+import psycopg2
+
+from qgis.core import (
+    QgsProject,
+    QgsDataSourceUri,
+    QgsMapLayer,
+    Qgis,
+    QgsVectorLayer,
+)
+from qgis.gui import QgsFileWidget, QgisInterface
+from qgis.PyQt import uic
+from qgis.PyQt.QtWidgets import QWizard, QLineEdit
+
+
+base_dir = os.path.dirname(__file__)
+ui_select_dbsync_page, base_select_dbsync_page = uic.loadUiType(
+    os.path.join(base_dir, "ui", "ui_switch_datasources_select_dbsync.ui")
+)
+ui_select_qgsproject_page, base_select_qgsproject_page = uic.loadUiType(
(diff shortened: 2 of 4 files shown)


Mergin/plugin.py
```diff
@@ -324,6 +331,38 @@ def configure_db_sync(self):
...
+        if not project_path:
+            iface.messageBar().pushMessage("Mergin", "Project is not saved, please save project first", Qgis.Warning)
+            return
+
+        if not check_mergin_subdirs(project_path):
+            iface.messageBar().pushMessage(
+                "Mergin", "Current project is not a Mergin project. Please open a Mergin project first.", Qgis.Warning
+            )
+            return
+
```
wonder-sk, line 345: worth adding a `unsaved_project_check()` test here as well - like in view_local_changes()

Mergin/switch_sources_dialog.py
```diff
@@ -0,0 +1,202 @@
...
+ui_select_qgsproject_page, base_select_qgsproject_page = uic.loadUiType(
+    os.path.join(base_dir, "ui", "ui_switch_datasources_select_updated_project.ui")
+)
+
+
+DBSYNC_PAGE = 0
+QGS_PROJECT_PAGE = 1
+
+
+@dataclass
```
wonder-sk, line 34: to me a dataclass is a class encapsulating some simple data - but a connection object isn't quite different IMHO. Could you please change it to ordinary class? (I also feel uneasy that you can currently pass e.g. `db_tables` to its `__init__` constructor...)

Mergin/switch_sources_dialog.py
```diff
@@ -0,0 +1,202 @@
+import os
+import yaml
+from dataclasses import dataclass, field
+import typing
+import re
+
+import psycopg2
```
wonder-sk, line 7: Hmm I am not sure if all QGIS installs will have psycopg2 automatically installed. May be good to handle importerror here, just to make sure that missing psycopg will not make the whole plugin unusable...

Mergin/switch_sources_dialog.py
```diff
@@ -0,0 +1,202 @@
+import os
+import yaml
```
wonder-sk, line 2: don't we need to add a wheel for this to the deps?

Mergin/switch_sources_dialog.py
```diff
@@ -0,0 +1,202 @@
...
+    db_tables: typing.List[str] = field(init=False)
+
+    def __post_init__(self):
+        try:
+            conn = psycopg2.connect(self.db_connection_info)
+        except Exception as e:
+            return
+        cur = conn.cursor()
+        cur.execute(f"SELECT tablename FROM pg_catalog.pg_tables WHERE schemaname = '{self.db_schema}'")
+        self.db_tables = [x[0] for x in cur.fetchall()]
```
wonder-sk, line 51: if db_tables is only used in convert_to_postgresql_layer(), do we need to have it as a member variable?

Mergin/switch_sources_dialog.py
```diff
@@ -0,0 +1,202 @@
...
+                )
+        else:
+            self.connections = []
+
+    def new_project_parent_folder(self) -> str:
+        return os.path.dirname(self.qgsproject_page.field("qgis_project"))
+
+    def convert_gpkg_layers_to_postgis_sources(self, result_qgsproject_path: str):
+        update_project = QgsProject()
+        update_project.read(self.qgis_project.fileName())
```
wonder-sk, line 186: I believe we don't need to read the full project - we don't need to resolve the layers, it will be much faster to read it

Mergin/switch_sources_dialog.py
```diff
@@ -0,0 +1,202 @@
...
+    def __init__(self, iface: QgisInterface, parent=None):
+        """Create a wizard"""
+        super().__init__(parent)
+
+        self.iface = iface
+        self.setWindowTitle("Create project with layers from Postgre Database")
+
+        self.connections: typing.List[Connection] = []
+
+        self.qgis_project = QgsProject.instance()
```
wonder-sk, line 141: let's not store reference to QgsProject.instance() to self.qgis_project - simply use QgsProject.instance() where needed, it will be easier to read the code

Mergin/switch_sources_dialog.py
```diff
@@ -0,0 +1,202 @@
...
+
+        layer: QgsMapLayer
+
+        for layer_id in project_layers:
+            layer = update_project.mapLayer(layer_id)
+
+            for dbsync_connection in self.connections:
+                if (
+                    layer.dataProvider().name() == "ogr"
+                    and dbsync_connection.sync_file in layer.dataProvider().dataSourceUri()
```
wonder-sk, line 198: shouldn't we do a better check for the URI match? e.g. if the sync_file is "data.gpkg" and there's also "backup/data.gpkg" file in the project, or maybe the QGIS project is also referencing a file outside of the MM project directory?

Mergin/switch_sources_dialog.py
```diff
@@ -0,0 +1,202 @@
...
+                    level=Qgis.Critical,
+                    duration=0,
+                )
+        else:
+            self.connections = []
+
+    def new_project_parent_folder(self) -> str:
+        return os.path.dirname(self.qgsproject_page.field("qgis_project"))
+
+    def convert_gpkg_layers_to_postgis_sources(self, result_qgsproject_path: str):
```
wonder-sk, line 184: I would prefer to have all the conversion code (+Connection class) outside of the GUI code, in a separate python file, so we can possibly also run auto-tests checking the conversion functionality...

Mergin/switch_sources_dialog.py
```diff
@@ -0,0 +1,202 @@
...
+            return
+        cur = conn.cursor()
+        cur.execute(f"SELECT tablename FROM pg_catalog.pg_tables WHERE schemaname = '{self.db_schema}'")
+        self.db_tables = [x[0] for x in cur.fetchall()]
+        self.valid = True
+
+    def convert_to_postgresql_layer(self, gpkg_layer: QgsVectorLayer) -> None:
+        layer_uri = gpkg_layer.dataProvider().dataSourceUri()
+
+        extract = re.search("\|layername=(.+)", layer_uri)
```
wonder-sk, line 57: try using QgsProviderRegistry.instance().decodeUri() instead of using regexps

Mergin/switch_sources_dialog.py
```diff
@@ -0,0 +1,202 @@
...
+    def convert_to_gpkg_layer(self, postgresql_layer: QgsVectorLayer, gpkg_folder: str) -> None:
+        gpkg = QgsVectorLayer(f"{gpkg_folder}/{self.sync_file}", "temp", "ogr")
+        gpkg_layers = [x.split("!!::!!")[1] for x in gpkg.dataProvider().subLayers()]
+
+        table_name = postgresql_layer.dataProvider().uri().table()
+
+        if table_name in gpkg_layers:
+            uri = f"{gpkg_folder}/{self.sync_file}|layername={table_name}"
+
+            postgresql_layer.setDataSource(uri, postgresql_layer.name(), "ogr")
```
wonder-sk, line 78: do we want to keep this code if it is not used / untested?

Mergin/switch_sources_dialog.py
```diff
@@ -0,0 +1,202 @@
...
+
+class ProjectUsePostgreConfigWizard(QWizard):
+    """Wizard for changing project layer sources from GPKG to Postgre DB."""
+
+    def __init__(self, iface: QgisInterface, parent=None):
+        """Create a wizard"""
+        super().__init__(parent)
+
+        self.iface = iface
+        self.setWindowTitle("Create project with layers from Postgre Database")
```
wonder-sk, line 137: I would suggest to use "PostgreSQL" instead of "Postgre" or "Postgre Database" everywhere

Mergin/plugin.py
```diff
@@ -324,6 +331,38 @@ def configure_db_sync(self):
...
+        dialog = ProjectUsePostgreConfigWizard(self.iface)
+        if not dialog.exec():
+            return
+
+        if check_mergin_subdirs(dialog.new_project_parent_folder()):
+            iface.messageBar().pushMessage(
+                "Mergin",
+                "The updated project should not be saved within Mergin directory. Please move it elsewhere.",
+                Qgis.Warning,
+            )
```
wonder-sk, line 364: I think this is a bit too late to say this :-) we should check that in the wizard dialog's code after picking output filename, and ask for a different output filename immediately

Mergin/ui/ui_switch_datasources_select_updated_project.ui
```diff
@@ -0,0 +1,68 @@
...
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>WizardPage</string>
+  </property>
+  <layout class="QVBoxLayout" name="verticalLayout">
+   <item>
+    <widget class="QLabel" name="label">
+     <property name="text">
+      <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;The resulting QGIS project should be stored outside of the Mergin project folder.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
wonder-sk, line 20: This probably should not need to be written there, and we would simply show a warning message box if the user tries to store it in a MM project directory...

Mergin/switch_sources_dialog.py
```diff
@@ -0,0 +1,202 @@
...
+
+    def nextId(self):
+        return QGS_PROJECT_PAGE
+
+
+class QgsProjectSelectionPage(ui_select_qgsproject_page, base_select_qgsproject_page):
+    """Wizard page with selection od QgsProject file selector."""
+
+    selectQgisProject: QgsFileWidget
+    lqgsproject_file: QLineEdit
```
wonder-sk, line 110: unusual naming with "l" prefix, and snake_case when other bits use mixedCase

wonder-sk (commented):
Could you please add screenshots to the PR description? Thanks!

tomasMizera:
After discussion with @wonder-sk we decided to park this one for now. We might revisit it in the future.

PR #542, opened 2023-11-07
do not show local mbtiles as not available offline (fix #540)

### LINKED ISSUE #540: Plugin flags the packaged mbtiles incorrectly as not available offline
When I try to sync, I have a bunch of mbtiles files for offline use, but the plugin flags them along the online layers and warns me that they would not be available offline:
image

alexbruy:
Raster  MBTiles are loaded with WMS provider and we need additional check whether this layer comes from local or remote source.

Fixes #540.

### DIFF
--- a/Mergin/validation.py
+++ b/Mergin/validation.py
@@ -3,7 +3,14 @@ import re
 from enum import Enum
 from collections import defaultdict
 
-from qgis.core import QgsMapLayerType, QgsProject, QgsVectorDataProvider, QgsExpression, QgsRenderContext
+from qgis.core import (
+    QgsMapLayerType,
+    QgsProject,
+    QgsVectorDataProvider,
+    QgsExpression,
+    QgsRenderContext,
+    QgsProviderRegistry,
+)
 
 from .help import MerginHelp
 from .utils import (
@@ -193,7 +200,15 @@ class MerginProjectValidator(object):
                 # might be vector tiles - no provider name
                 continue
             if dp_name in QGIS_NET_PROVIDERS + QGIS_DB_PROVIDERS:
-                w.items.append(layer.name())
+                # raster mbtiles are loaded using WMS provider, so we need to
+                # check whether this is a local file or remote service
+                if dp_name.lower() == "wms":
+                    uri = QgsProviderRegistry.instance().decodeUri("wms", layer.source())
+                    is_local = os.path.isfile(uri["path"]) if "path" in uri else False
+                    if not is_local:
+                        w.items.append(layer.name())
+                else:
+                    w.items.append(layer.name())
 
         if w.items:
             self.issues.append(w)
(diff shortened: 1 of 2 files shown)


Mergin/validation.py
```diff
@@ -193,7 +200,15 @@ def check_offline(self):
...
-                w.items.append(layer.name())
+                # raster mbtiles are loaded using WMS provider, so we need to
+                # check whether this is a local file or remote service
+                if dp_name.lower() == "wms":
+                    uri = QgsProviderRegistry.instance().decodeUri("wms", layer.source())
+                    is_local = os.path.isfile(uri["path"]) if "path" in uri else False
+                    if not is_local:
+                        w.items.append(layer.name())
+                else:
+                    w.items.append(layer.name())
```
wonder-sk, line 211: I think we should handle also vector tiles MBTiles here... from the code it looks like that vector tiles MBTiles layer would always get flagged as not being offline, or never get flagged - depending on whether the data provider is present. For vector tiles, we should rather check the layer type, and then check if URI contains `type=mbtiles` (and while at it, we could also allow `type=vtpk` as offline use)

wonder-sk (commented):
This should handle raster MBTiles, but

Mergin/validation.py
alexbruy: Should be addressed

wonder-sk: How about the code few lines above... isn't `continue` going to skip the online/offline checks?
```python
            try:
                dp_name = layer.dataProvider().name()
            except AttributeError:
                # might be vector tiles - no provider name
                continue
```

wonder-sk: They do... in QGIS >= 3.22... but in earlier versions (since introduction of vector tiles in 3.14) they did not have providers assigned.

Mergin/validation.py
```diff
@@ -191,9 +198,16 @@ def check_offline(self):
                 dp_name = layer.dataProvider().name()
             except AttributeError:
                 # might be vector tiles - no provider name
-                continue
+                if layer.type() == QgsMapLayerType.VectorTileLayer:
+                    if "type=mbtiles" not in layer.source() and "type=vtpk" not in layer.source():
+                        w.items.append(layer.name())
```
wonder-sk, line 203: I think on QGIS < 3.22 things will not work because `dp_name` will be undefined on the next line.

I would suggest to restructure the whole section for better clarity e.g. like this (not tested):
```python
# special check for vector tile layers because in QGIS < 3.22 they may not have data provider assigned
if layer.type() == QgsMapLayerType.VectorTileLayer:
    # mbtiles/vtpk are always local files
    if "type=mbtiles" in layer.source() or "type=vtpk" in layer.source():
        continue
    w.items.append(layer.name())
    continue

dp_name = layer.dataProvider().name()
if dp_name in QGIS_NET_PROVIDERS + QGIS_DB_PROVIDERS:
    # raster tiles in mbtiles are always local files
    if dp_name == "xyz" and "type=mbtiles" in layer.source():
        continue
    w.items.append(layer.name())
```

PR #546, opened 2023-11-09
Add dock with project history

alexbruy:
Add a dock widget with project history. History is loaded from the server in the background thread when dock opens for the first time, then on each sync history is updated with the new information. Local project version is shown in bold.

Versions in the dock have a context menu with two actions:
 - viewing version metadata (author, project size, creation date, user agent, etc) and details (changed tables, number of added/removed/modified records).
 - viewing version changes in the diff viewer dialog. Plugin downloads all necessary changesets in the background thread and then uses them to create a diff. Downloaded files are cached and local files will be used when diff is requested again.

Peek 2023-11-09 12-36

### DIFF
--- a/Mergin/diff.py
+++ b/Mergin/diff.py
@@ -3,6 +3,7 @@ import json
 import base64
 import sqlite3
 import tempfile
+import glob
 
 from qgis.PyQt.QtCore import QVariant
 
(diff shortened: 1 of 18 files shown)


ValentinBuira:
The work will continue in the branch project-history-refactor and a different PR

PR #577, opened 2024-03-15
strip trailing whitespace from project name

### LINKED ISSUE #552: plugin does not trim whitespace in the new project name
In plugin I can create a new project with some whitespaces at the end. Server uses strip() to remove leading and trailing spaces but plugin creates a folder and files with the spaces. Server cannot find the project files.

How to reproduce
Create a new project with trailing spaces in plugin

What I expect
Plugin should also trim the whitespaces rather than write the name is invalid.

---
Author of the original task:
Herman Snevajs

Original tags: 
migrate to github

Original task URL:  https://app.clickup.com/t/861mynfjj

JanCaha:
Strip spaces from Project Name in Create new project window.

Fix #552

### DIFF
--- a/Mergin/create_project_wizard.py
+++ b/Mergin/create_project_wizard.py
@@ -174,7 +174,7 @@ class ProjectSettingsPage(ui_proj_settings, base_proj_settings):
         if not self.project_owner_cbo.currentData(Qt.UserRole):
             self.create_warning("You do not have permissions to create a project in this workspace!")
             return
-        proj_name = self.project_name_ledit.text()
+        proj_name = self.project_name_ledit.text().strip()
         if not proj_name:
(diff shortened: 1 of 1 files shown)


ValentinBuira (approved):

alexd0793:
Error on qgis plugin when trying to create a project with spaces at the beginning and at the end of it's name 
image

ValentinBuira (changes requested):
This fail to create the project:
* On windows see  comment from @alexd0793 on the PR
*  On server side : 
image
(actually the project is created on the server side but is not accessible from the plugin)

Do you think you could have a look @JanCaha ?

ValentinBuira:
Thanks @JanCaha , looking good now merging then :-)

PR #601, opened 2024-07-09
Better dialog error when trying to create a project with an already existing name and the projects limit hit error.

### LINKED ISSUE #561: Better notification for already runing synchronization
When on the mobile app I have a longer synchronization (for example with a bigger file) and then during it, I synchronize with a plugin, it shows this error. The error is correct but nicer notification would be better. For example, don't show 400 bad requests. 

Screenshot at 2024-01-23 09-07-16

### LINKED ISSUE #599: Better dialog for error "project with the same name already exists"
At the moment we show a generic dialog with the entire error message from the server. Let's add a new dialog instead to show tha info nicely.

Relates to https://github.com/MerginMaps/qgis-plugin/issues/561

ValentinBuira:
This PR improve the dialog error given to user when trying to create a project and sync project. Instead of giving a dump of the server error we display only the part that matter to the users

So far we handle 
* Project with an already existing name 
* projects limit hit error
* Storage limit error
* Another syncing process running

This PR  also relies on another PR in the mergin python api client (see : MerginMaps/python-api-client/pull/209 )

resolve #599 
resolve #561

Before: 
Capture d’écran before
<br>
Capture d’écran before quota

After: 
Screenshot from 2024-07-09 15-27-29

Screenshot from 2024-07-09 16-11-00

### DIFF
--- a/Mergin/projects_manager.py
+++ b/Mergin/projects_manager.py
@@ -86,6 +86,19 @@ class MerginProjectsManager(object):
         QApplication.setOverrideCursor(Qt.WaitCursor)
         try:
             self.mc.create_project(full_project_name, is_public)
+        except ClientError as e:
+            QApplication.restoreOverrideCursor()
+            msg= str(e)
+            #User friendly error messages
+            if e.http_error == 409:
+                msg = f"Project with the name {full_project_name} already exists\n Try renaming the project"
+            if e.http_error == 422 and e.server_code == 'ProjectsLimitHit':
+                msg =  ("Maximum number of projects is reached. Please upgrade your subscription to create new projects\n"
+                        f"Maximum of projects: {e.server_response["projects_quota"]}"
+                )
+
+            QMessageBox.critical(None, "Create Project", "Failed to create Mergin Maps project.\n" + msg)
+            return False
         except Exception as e:
             QApplication.restoreOverrideCursor()
             QMessageBox.critical(None, "Create Project", "Failed to create Mergin Maps project.\n" + str(e))
--- a/Mergin/projects_manager.py
+++ b/Mergin/projects_manager.py
@@ -12,6 +12,7 @@ from urllib.error import URLError
 from .sync_dialog import SyncDialog
 from .utils import (
     ClientError,
+    ErrorCode,
     InvalidProject,
     get_local_mergin_projects_info,
     LoginError,
@@ -91,8 +92,8 @@ class MerginProjectsManager(object):
             msg = str(e)
             # User friendly error messages
             if e.http_error == 409:
-                msg = f"Project with the name \"{project_name}\" already exists in the workspace \"{namespace}\"\n Try renaming the project"
-            if e.http_error == 422 and e.server_code == "ProjectsLimitHit":
+                msg = f'Project with the name "{project_name}" already exists in the workspace "{namespace}"\n Try renaming the project'
+            if e.http_error == 422 and e.server_code == ErrorCode.ProjectsLimitHit.value:
                 msg = (
                     "Maximum number of projects is reached. Please upgrade your subscription to create new projects\n"
                     f"Maximum of projects: {e.server_response["projects_quota"]}"
--- a/Mergin/projects_manager.py
+++ b/Mergin/projects_manager.py
@@ -24,7 +24,7 @@ from .utils import (
     unsaved_project_check,
     UnsavedChangesStrategy,
     write_project_variables,
-    bytes_to_mb
+    bytes_to_human_size,
 )
 
 from .mergin.merginproject import MerginProject
@@ -100,7 +100,7 @@ class MerginProjectsManager(object):
                     f"Maximum of projects: {e.server_response["projects_quota"]}"
                 )
             elif e.server_code == ErrorCode.StorageLimitHit.value:
-                msg = f"{e.detail}]\nCurrent limit: {bytes_to_mb(dlg.exception.server_response["storage_limit"])} Mb"
+                msg = f"{e.detail}]\nCurrent limit: {bytes_to_human_size(dlg.exception.server_response["storage_limit"])}"
 
             QMessageBox.critical(None, "Create Project", "Failed to create Mergin Maps project.\n" + msg)
             return False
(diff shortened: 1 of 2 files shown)


Poulpator:
For reviewer, I think it works fine by now but as we handle more and more case we will have a lot of duplicate code because the same error can happen in mutiple place : for exemple StorageLimitHit happen both in `create_project` and `sync_project` 

I advice to make a more general framework to handle error dialog in the futur

Mergin/projects_manager.py
```diff
@@ -86,6 +87,22 @@ def create_project(self, project_name, project_dir, is_public, namespace):
...
+            # User friendly error messages
+            if e.http_error == 409:
+                msg = f'Project with the name "{project_name}" already exists in the workspace "{namespace}"\n Try renaming the project'
+            elif e.http_error == 422 and e.server_code == ErrorCode.ProjectsLimitHit.value:
+                msg = (
+                    "Maximum number of projects is reached. Please upgrade your subscription to create new projects\n"
+                    f"Maximum of projects: {e.server_response["projects_quota"]}"
+                )
+            elif e.http_error == 422 and e.server_code == ErrorCode.StorageLimitHit.value:
+                msg = f"{e.detail}]\nCurrent limit: {dlg.exception.server_response["storage_limit"]} Mb"
```
tomasMizera, line 102: Do we know if the `dlg.exception.server_reponse["storage_limit"]` always reports MBs?

Mergin/projects_manager.py
```diff
@@ -86,6 +87,22 @@ def create_project(self, project_name, project_dir, is_public, namespace):
         QApplication.setOverrideCursor(Qt.WaitCursor)
         try:
             self.mc.create_project(full_project_name, is_public)
+        except ClientError as e:
+            QApplication.restoreOverrideCursor()
+            msg = str(e)
+            # User friendly error messages
+            if e.http_error == 409:
+                msg = f'Project with the name "{project_name}" already exists in the workspace "{namespace}"\n Try renaming the project'
+            elif e.http_error == 422 and e.server_code == ErrorCode.ProjectsLimitHit.value:
```
tomasMizera, line 96: Checking only the error code should be enough and it is also better for the future if we'd like to change the http response code.

```suggestion
            elif e.server_code == ErrorCode.ProjectsLimitHit.value:
```

Poulpator: Fixed for this one and all others occurrences in latest commit

Mergin/projects_manager.py
ValentinBuira: Added the function  `bytes_to_mb` in utils.py

Mergin/projects_manager.py
```diff
@@ -99,7 +100,7 @@ def create_project(self, project_name, project_dir, is_public, namespace):
                     f"Maximum of projects: {e.server_response["projects_quota"]}"
                 )
             elif e.server_code == ErrorCode.StorageLimitHit.value:
-                msg = f"{e.detail}]\nCurrent limit: {dlg.exception.server_response["storage_limit"]} Mb"
+                msg = f"{e.detail}]\nCurrent limit: {bytes_to_mb(dlg.exception.server_response["storage_limit"])} Mb"
```
tomasMizera, line 103:
```suggestion
                msg = f"{e.detail}]\nCurrent limit: {bytes_to_mb(dlg.exception.server_response["storage_limit"])} MB"
```

Mergin/utils.py
```diff
@@ -1505,3 +1505,12 @@ def get_layer_by_path(path):
...
             return layer
+
+def bytes_to_mb(bytes : int):
+    """
+    Convert bytes to megabytes assuming base of bytes in binary
+    
+    Returns:
+    float: Result in megabytes
+    """
+    return bytes / (2**20)
```
tomasMizera, line 1516: Thinking about this again, it will be nicer if it could convert the input size into human readable output. Imagine you are on a team plan (with 50GBs), you would now see a max limit of 50000 MB, but we could show 50 GB instead to make it more readable.

There is a function like this in our mobile app already. Can you please add an equivalent of it?
https://github.com/MerginMaps/mobile/blob/052b5b663e3ad6095995f58adb16f40a82c302ae/core/coreutils.cpp#L314-L323

In fact, moving this function to `python-api-client` will be a good idea so that we can reuse it later in other places too.

Mergin/projects_manager.py
```diff
@@ -86,6 +88,24 @@ def create_project(self, project_name, project_dir, is_public, namespace):
...
+        except ClientError as e:
+            QApplication.restoreOverrideCursor()
+            msg = str(e)
+            # User friendly error messages
+            if e.http_error == 409:
+                msg = f'Project with the name "{project_name}" already exists in the workspace "{namespace}"\n Try renaming the project'
+            elif e.server_code == ErrorCode.ProjectsLimitHit.value:
+                msg = (
+                    "Maximum number of projects is reached. Please upgrade your subscription to create new projects\n"
+                    f"Maximum of projects: {e.server_response["projects_quota"]}"
```
tomasMizera, line 100: Minor text change

```python
"Maximum number of projects reached. Please upgrade your subscription to create new projects.\n"
f"Projects quota: {e.server_response['projects_quota']}"
```

Mergin/projects_manager.py
```diff
@@ -86,6 +88,24 @@ def create_project(self, project_name, project_dir, is_public, namespace):
         QApplication.setOverrideCursor(Qt.WaitCursor)
         try:
             self.mc.create_project(full_project_name, is_public)
+        except ClientError as e:
+            QApplication.restoreOverrideCursor()
+            msg = str(e)
+            # User friendly error messages
+            if e.http_error == 409:
+                msg = f'Project with the name "{project_name}" already exists in the workspace "{namespace}"\n Try renaming the project'
```
tomasMizera, line 96:
```suggestion
                msg = f'Project named "{project_name}" already exists in the workspace "{namespace}".\nPlease try renaming the project.'
```

Mergin/projects_manager.py
```diff
@@ -351,7 +371,14 @@ def sync_project(self, project_dir, project_name=None):
             if isinstance(dlg.exception, LoginError):
                 login_error_message(dlg.exception)
             elif isinstance(dlg.exception, ClientError):
-                QMessageBox.critical(None, "Project sync", "Client error: " + str(dlg.exception))
+                if dlg.exception.http_error == 400 and "Another process" in dlg.exception.detail:
+                    # To note we check for a string since error in flask doesn't return server error code
+                    msg = dlg.exception.detail
```
tomasMizera, line 376: What is the message from the server? On the mobile app we say "Somebody else is syncing..". Can we here say something similar like `Somebody else is syncing, please try again later.` please?

tomasMizera (approved):
Just some AI suggestions for the texts, otherwise it looks good 👍🏻

Mergin/projects_manager.py
ValentinBuira: The message from server is "Another Process is running. Please try later"

Agreed it could be more explicit

ValentinBuira: I change the error message to `Somebody else is syncing, please try again later` to match more closely the mobile app

alexd0793:
Testing all cases, looking good :)

PR #602, opened 2024-07-09
Sort workspaces list in new project wizard

erpas:
this is mainly for support agents who need to find the right one in a list of more than 100 of them

ws_list

### DIFF
--- a/Mergin/create_project_wizard.py
+++ b/Mergin/create_project_wizard.py
@@ -115,7 +115,7 @@ class ProjectSettingsPage(ui_proj_settings, base_proj_settings):
 
     def populate_namespace_cbo(self):
         if self.parent.workspaces is not None:
-            for ws in self.parent.workspaces:
+            for ws in sorted(self.parent.workspaces, key=lambda x: x["name"].lower()):
                 is_writable = ws.get("role", "owner") in ["owner", "admin", "writer"]
                 self.project_owner_cbo.addItem(ws["name"], is_writable)
 
--- a/scripts/deploy_win.py
+++ b/scripts/deploy_win.py
@@ -1,12 +1,15 @@
 import os
 import shutil
+
+profile = "default"
+
 this_dir = os.path.dirname(os.path.realpath(__file__))
 home_dir = os.path.expanduser("~")
-dest_dir_plug = os.path.join(home_dir, "AppData", "Roaming", "QGIS", "QGIS3", "profiles", "default", "python", "plugins", "Mergin")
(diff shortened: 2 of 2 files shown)


scripts/deploy_win.py
tomasMizera, line 1: Is this change intentional @erpas?

erpas: Yes, it was intentional. 

On Windows we often get `OSError` not only when the plugin directory does not exist but mainly because of files being in use and it is not possible to remove it.

Mergin/create_project_wizard.py
```diff
@@ -115,7 +115,7 @@ def initializePage(self):
 
     def populate_namespace_cbo(self):
         if self.parent.workspaces is not None:
-            for ws in self.parent.workspaces:
+            for ws in sorted(self.parent.workspaces, key=lambda x: x["name"].lower()):
```
Poulpator, line 118: Code wise look fine however the sort is currently alphanumeric where I think users will expect a natural sort

So for example if project names in a workspace look like this
```
project 0
project 1
project 2
project 3
project 4
project 5
project 6
project 7
project 8
project 9
project 10
project 11
project 12
```

it will be actually sorted as
```
project 0
project 1
project 10
project 11
project 12
project 2
project 3
project 4
project 5
project 6
project 7
project 8
project 9
```

I'm not sure how often users name their projects numerically but we can implement a natural in python like this https://stackoverflow.com/a/4836734

erpas: Thanks for pointing this out. However, as stated previously, this PR is mainly for support agents who deal with dozens of workspaces and we do not care whether the sorting is natural or not :-)

Normally, users have just a couple of them at most and I doubt their names have numbers.

Poulpator: Okay then it looks good to me  :+1:

scripts/deploy_win.py
```diff
@@ -1,12 +1,15 @@
...
 home_dir = os.path.expanduser("~")
-dest_dir_plug = os.path.join(home_dir, "AppData", "Roaming", "QGIS", "QGIS3", "profiles", "default", "python", "plugins", "Mergin")
+dest_dir_plug = os.path.join(home_dir, "AppData", "Roaming", "QGIS", "QGIS3", "profiles", profile, "python", "plugins", "Mergin")
 print(dest_dir_plug)
 src_dir_plug = os.path.join(os.path.dirname(this_dir), "Mergin")
 try:
     shutil.rmtree(dest_dir_plug)
 except OSError:
-    pass  # directory doesn't not exist
+    print("Could not remove Mergin")
```
tomasMizera, line 14: Let's make PP proud

```suggestion
    print("Could not remove Mergin Maps")
```

tomasMizera (approved):

PR #611, opened 2024-09-05
Fix CI tests

tomasMizera:
Latest docker image has some issues

### DIFF
--- a/.github/workflows/run-test.yml
+++ b/.github/workflows/run-test.yml
@@ -18,7 +18,7 @@ jobs:
 
     strategy:
       matrix:
-        docker_tags: [release-3_22, latest]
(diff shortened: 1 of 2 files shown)


ValentinBuira:
Is it worth looking into fixing the docker image in QGIS  upstream, we might have too either way on next QGS ltr version ?  Otherwise look good to me

tomasMizera:
> Is it worth looking into fixing the docker image in QGIS upstream, we might have too either way on next QGS ltr version ? Otherwise look good to me

Sure, but from the Slack message (https://lutraconsulting.slack.com/archives/CC95JB2UF/p1725525505962069), I understand it wouldn't be that easy.

PR #615, opened 2024-09-12
Revert actions to @v3

tomasMizera:
To (hopefully) fix the CI. If it won't work, let's revert back to version 3 and kill dependabot 👼🏻 

EDIT
- Reverted back to`v3`, there are several braking changes in the `v4`. We will eventually migrate, but let's keep it for later, see https://github.com/actions/upload-artifact?tab=readme-ov-file#v4---whats-new


ValentinBuira:
Yep I'll do a separated PR once tests are back working

PR #621, opened 2024-09-24
Enable packaging and test workflows to be triggered mannually with custom inputs  

ValentinBuira:
This PR enable a workflow to be triggered manually with custom inputs using `workflow_dispatch` event.
The packages.yml and run-test.yml workflow are currently supported 

This PR reuse the same workflow used when a commit is  pushed and should remain compatible 

Currently we support the following argument(inputs)
* `PYTHON_API_CLIENT_VER` : For python-api-client you can choice any release, tag or branch
* `GEODIFF_VER`: For geodiff only the version uploaded on  PyPi are supported  

To note, you  can't trigger the workflow from the github web interface since it's not yet on the default branch, to test from the  cli interface, use the following command e.g:

For packaging: 
```
gh workflow run "Build Mergin Plugin Packages" --ref workflow-dispatch -f PYTHON_API_CLIENT_VER="master"
```
For test:
```
gh workflow run "Run Mergin Plugin Tests" --ref workflow-dispatch -f PYTHON_API_CLIENT_VER="project-history"
```

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -3,19 +3,19 @@ on:
   push:
   workflow_dispatch:
     inputs:
-      API_CLIENT_VER:
-        description: 'Mergin client version either a tag, release, or branch'
+      PYTHON_API_CLIENT_VER:
+        description: 'python-api-client version: either a tag, release, or a branch'
         required: true
         default: 'master'
         type: string
-      LIB_GEODIFF_VER:
+      GEODIFF_VER:
         description: 'Geodiff version released on PyPI repository'
         default: '2.0.4'
         type: string
 env:
   # Assign the version provided by 'workflow_dispatch' if available; otherwise, use the default.
-  MERGIN_CLIENT_VER: ${{ inputs.API_CLIENT_VER != '' && inputs.API_CLIENT_VER || '0.9.3' }}
-  GEODIFF_VER:  ${{ inputs.LIB_GEODIFF_VER != '' && inputs.LIB_GEODIFF_VER || '2.0.4' }}
+  PYTHON_API_CLIENT_VER: ${{ inputs.PYTHON_API_CLIENT_VER != '' && inputs.PYTHON_API_CLIENT_VER || '0.9.3' }}
+  GEODIFF_VER:  ${{ inputs.GEODIFF_VER != '' && inputs.GEODIFF_VER || '2.0.4' }}
   PYTHON_VER: "38"
   PLUGIN_NAME: Mergin  
 jobs:
@@ -120,7 +120,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           repository: MerginMaps/mergin-py-client
-          ref: ${{ env.MERGIN_CLIENT_VER }}
+          ref: ${{ env.PYTHON_API_CLIENT_VER }}
           path: mergin-py-client
 
       - name: prepare py-client dependencies
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -3,19 +3,19 @@ on:
   push:
   workflow_dispatch:
     inputs:
-      MERGIN_CLIENT_VER:
+      API_CLIENT_VER:
         description: 'Mergin client version either a tag, release, or branch'
         required: true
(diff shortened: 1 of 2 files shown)


.github/workflows/packages.yml
```diff
@@ -1,11 +1,23 @@
...
+        type: string
+      GEODIFF_VER:
+        description: 'Geodiff version released on PyPI repository'
+        required: true
+        default: '2.0.4'
+        type: string
 env:
-  MERGIN_CLIENT_VER: "0.9.3"
-  GEODIFF_VER: "2.0.4"
+  MERGIN_CLIENT_VER: ${{ inputs.MERGIN_CLIENT_VER != '' && inputs.MERGIN_CLIENT_VER|| '0.9.3' }}
```
tomasMizera, line 17: Please add explanation comments

.github/workflows/packages.yml
```diff
@@ -1,11 +1,24 @@
...
+      MERGIN_CLIENT_VER:
+        description: 'Mergin client version either a tag, release, or branch'
+        required: true
+        default: 'master'
+        type: string
+      GEODIFF_VER:
+        description: 'Geodiff version released on PyPI repository'
+        required: true
+        default: '2.0.4'
+        type: string
```
tomasMizera, line 15: I'd probably mark the geodiff as not required, not sure how often we'd need to do that :)

.github/workflows/packages.yml
```diff
@@ -1,11 +1,24 @@
...
 env:
-  MERGIN_CLIENT_VER: "0.9.3"
-  GEODIFF_VER: "2.0.4"
+  # Assign the version provided by 'workflow_dispatch' if available; otherwise, use the default.
+  MERGIN_CLIENT_VER: ${{ inputs.MERGIN_CLIENT_VER != '' && inputs.MERGIN_CLIENT_VER|| '0.9.3' }}
+  GEODIFF_VER:  ${{ inputs.GEODIFF_VER != '' && inputs.GEODIFF_VER|| '2.0.4' }}
```
tomasMizera, line 19: I know bash can be picky when it comes to spaces, not sure if this matters or not
```suggestion
  GEODIFF_VER:  ${{ inputs.GEODIFF_VER != '' && inputs.GEODIFF_VER || '2.0.4' }}
```

.github/workflows/run-test.yml
```diff
@@ -1,9 +1,18 @@
 name: Run Mergin Plugin Tests
 
-on: [push]
+on: 
+  push:
+  workflow_dispatch:
+    inputs:
+      MERGIN_CLIENT_VER:
```
tomasMizera, line 7: Let's not introduce old namings :) 
```suggestion
      API_CLIENT_VER:
```

tomasMizera: The same below "Mergin client version either ..."

tomasMizera (changes requested):
Small changes, otherwise all looks good :)

.github/workflows/packages.yml
ValentinBuira: Added the comment:  
```
 # Assign the version provided by 'workflow_dispatch' if available; otherwise, use the default.
```

.github/workflows/packages.yml
ValentinBuira: Added the comment:  
```
 # Assign the version provided by 'workflow_dispatch' if available; otherwise, use the default.
```

.github/workflows/packages.yml
ValentinBuira: You right not used much, removed required in latest

.github/workflows/run-test.yml
ValentinBuira: Renamed all the occurences I settled for `PYTHON_API_CLIENT_VER` in the end

.github/workflows/packages.yml
ValentinBuira: Added the space between operator just in case ;-)

tomasMizera (approved):
Let's see how it works on master! :)

tomasMizera:
Thanks for the adjustments!

PR #628, opened 2024-10-02
Fix unreported : Fix crash with f string expression

ValentinBuira:
In some older python version, f string are confused when using the string delimiter inside the expression. That makes crash the plugin on installs

This PR fix the issues by using a mix of simple and double quote


VitorVieiraZ (approved):
All working well now.

PR #630, opened 2024-10-04
Project history : Add a version viewer windows in the plugin

ValentinBuira:
This PR is a refactor and enhancement to bring a new history window to browse the history of a project directly from the plugin

To open the new windows a new toolbar button has been added.

The windows is divided in three parts for the users to be able to:
* Scroll thought projects versions the left
* Explore change visually in case the version is diffable 
* Viewing layer changed, version metadata (author creation date, user agent, etc...), and changesets details .

This PR is marked as draft because some testing are still required and the feature still have rough edges

Screenshot from 2024-10-04 10-51-42

For reviewer a brief technical documentation is available in the comment

### DIFF
--- a/Mergin/diff.py
+++ b/Mergin/diff.py
@@ -7,7 +7,7 @@ import glob
 
 from qgis.PyQt.QtCore import QVariant
 
-from qgis.PyQt.QtGui import QColor, QIcon
+from qgis.PyQt.QtGui import QColor
 
 from qgis.core import (
     QgsApplication,
@@ -530,17 +530,3 @@ def style_diff_layer(layer, schema_table):
         )
         r = QgsRuleBasedRenderer(root_rule)
         layer.setRenderer(r)
-
-
-def icon_for_layer(layer) -> QIcon:
-    geom_type = layer.geometryType()
-    if geom_type == QgsWkbTypes.PointGeometry:
-        return QgsApplication.getThemeIcon("/mIconPointLayer.svg")
-    elif geom_type == QgsWkbTypes.LineGeometry:
-        return QgsApplication.getThemeIcon("/mIconLineLayer.svg")
-    elif geom_type == QgsWkbTypes.PolygonGeometry:
-        return QgsApplication.getThemeIcon("/mIconPolygonLayer.svg")
-    elif geom_type == QgsWkbTypes.UnknownGeometry:
-        return QgsApplication.getThemeIcon("/mIconGeometryCollectionLayer.svg")
-    else:
-        return QgsApplication.getThemeIcon("/mIconTableLayer.svg")
--- a/Mergin/diff.py
+++ b/Mergin/diff.py
@@ -330,7 +330,7 @@ def make_version_changes_layers(project_path, version):
 
     layers = []
     version_dir = os.path.join(project_path, ".mergin", ".cache", f"v{version}")
-    for f in glob.iglob("*.gpkg", root_dir=version_dir):
+    for f in glob.iglob(f"{version_dir}/*.gpkg"):
         gpkg_file = os.path.join(version_dir, f)
         schema_file = gpkg_file + "-schema.json"
         if not os.path.exists(schema_file):
@@ -348,6 +348,10 @@ def make_version_changes_layers(project_path, version):
         diff = parse_diff(geodiff, changeset_file)
 
         for table_name in diff.keys():
+            if table_name.startswith("gpkg_"):
+                # db schema reported by geodiff exclude layer named "gpkg_*" 
+                # We skip it in the layer displayed to the user  
+                continue
             fields, cols_to_fields = create_field_list(db_schema[table_name])
             geom_type, geom_crs = get_layer_geometry_info(schema_json, table_name)
 
@@ -377,7 +381,7 @@ def make_version_changes_layers(project_path, version):
 
 def find_changeset_file(file_name, version_dir):
     """Returns path to the diff file for the given version file"""
-    for f in glob.iglob("*.gpkg-diff*", root_dir=version_dir):
+    for f in glob.iglob(f"{version_dir}/*.gpkg-diff*"):
         if f.startswith(file_name):
             return os.path.join(version_dir, f)
     return None
--- a/Mergin/diff.py
+++ b/Mergin/diff.py
@@ -330,7 +330,7 @@ def make_version_changes_layers(project_path, version):
 
     layers = []
     version_dir = os.path.join(project_path, ".mergin", ".cache", f"v{version}")
-    for f in glob.iglob("*.gpkg", root_dir=version_dir):
+    for f in glob.iglob(f"{version_dir}/*.gpkg"):
         gpkg_file = os.path.join(version_dir, f)
         schema_file = gpkg_file + "-schema.json"
         if not os.path.exists(schema_file):
@@ -377,7 +377,7 @@ def make_version_changes_layers(project_path, version):
 
 def find_changeset_file(file_name, version_dir):
     """Returns path to the diff file for the given version file"""
-    for f in glob.iglob("*.gpkg-diff*", root_dir=version_dir):
+    for f in glob.iglob(f"{version_dir}/*.gpkg-diff*"):
         if f.startswith(file_name):
             return os.path.join(version_dir, f)
     return None
--- a/Mergin/diff.py
+++ b/Mergin/diff.py
@@ -361,7 +361,7 @@ def make_version_changes_layers(project_path, version):
                 continue
 
             uri = f"{geom_type}?crs=epsg:{geom_crs}" if geom_crs else geom_type
-            vl = QgsVectorLayer(uri, f"[v{version} changes] {table_name}", "memory")
+            vl = QgsVectorLayer(uri, table_name, "memory")
             if not vl.isValid():
                 continue
 
--- a/Mergin/diff.py
+++ b/Mergin/diff.py
@@ -220,6 +220,8 @@ def diff_table_to_features(diff_table, schema_table, fields, cols_to_flds, db_co
 
             for i in range(len(db_row)):
                 if i == geom_col_index:
+                    if db_row[i] == None:
+                        continue
                     wkb = parse_gpkg_geom_encoding(db_row[i])
                     g = QgsGeometry()
                     g.fromWkb(wkb)
--- a/Mergin/diff.py
+++ b/Mergin/diff.py
@@ -242,6 +242,9 @@ def diff_table_to_features(diff_table, schema_table, fields, cols_to_flds, db_co
                 value = "?"
 
             if i == geom_col_index:
+                if value == None:
+                    # Empty geometry
+                    continue
                 wkb_with_gpkg_hdr = base64.decodebytes(value.encode("ascii"))
                 wkb = parse_gpkg_geom_encoding(wkb_with_gpkg_hdr)
                 g = QgsGeometry()
--- a/Mergin/diff.py
+++ b/Mergin/diff.py
@@ -330,7 +330,7 @@ def make_version_changes_layers(project_path, version):
 
     layers = []
     version_dir = os.path.join(project_path, ".mergin", ".cache", f"v{version}")
-    for f in glob.iglob(f"{version_dir}/*.gpkg"):
+    for f in glob.iglob("*.gpkg", root_dir=version_dir):
         gpkg_file = os.path.join(version_dir, f)
         schema_file = gpkg_file + "-schema.json"
         if not os.path.exists(schema_file):
@@ -377,7 +377,7 @@ def make_version_changes_layers(project_path, version):
 
 def find_changeset_file(file_name, version_dir):
     """Returns path to the diff file for the given version file"""
-    for f in glob.iglob(f"{version_dir}/*.gpkg-diff*"):
+    for f in glob.iglob("*.gpkg-diff*", root_dir=version_dir):
         if f.startswith(file_name):
             return os.path.join(version_dir, f)
     return None
--- a/Mergin/diff.py
+++ b/Mergin/diff.py
@@ -349,8 +349,8 @@ def make_version_changes_layers(project_path, version):
 
         for table_name in diff.keys():
             if table_name.startswith("gpkg_"):
-                # db schema reported by geodiff exclude layer named "gpkg_*" 
-                # We skip it in the layer displayed to the user  
+                # db schema reported by geodiff exclude layer named "gpkg_*"
+                # We skip it in the layer displayed to the user
                 continue
             fields, cols_to_fields = create_field_list(db_schema[table_name])
             geom_type, geom_crs = get_layer_geometry_info(schema_json, table_name)
--- a/Mergin/utils.py
+++ b/Mergin/utils.py
@@ -16,7 +16,7 @@ import re
 
 from qgis.PyQt.QtCore import QSettings, QVariant
 from qgis.PyQt.QtWidgets import QMessageBox, QFileDialog
-from qgis.PyQt.QtGui import QPalette, QColor
+from qgis.PyQt.QtGui import QPalette, QColor, QIcon
 from qgis.core import (
     NULL,
     Qgis,
@@ -1573,3 +1573,18 @@ def parse_user_agent(user_agent: str) -> str:
         return "Mergin Maps Python Client"
     else:  # For uncommon user agent we display user agent as is
         return user_agent
+
+
+def icon_for_layer(layer) -> QIcon:
+    # Used in diff viewer and history viewer
+    geom_type = layer.geometryType()
+    if geom_type == QgsWkbTypes.PointGeometry:
+        return QgsApplication.getThemeIcon("/mIconPointLayer.svg")
+    elif geom_type == QgsWkbTypes.LineGeometry:
+        return QgsApplication.getThemeIcon("/mIconLineLayer.svg")
+    elif geom_type == QgsWkbTypes.PolygonGeometry:
+        return QgsApplication.getThemeIcon("/mIconPolygonLayer.svg")
+    elif geom_type == QgsWkbTypes.UnknownGeometry:
+        return QgsApplication.getThemeIcon("/mIconGeometryCollectionLayer.svg")
+    else:
+        return QgsApplication.getThemeIcon("/mIconTableLayer.svg")
--- a/Mergin/utils.py
+++ b/Mergin/utils.py
@@ -1557,7 +1557,7 @@ def format_datetime(date_string):
 
 def parse_user_agent(user_agent: str) -> str:
     browsers = ["Chrome", "Firefox", "Mozilla", "Opera", "Safari", "Webkit"]
-    if any([ browser in user_agent for  browser in browsers]):
+    if any([browser in user_agent for browser in browsers]):
         return "Web dashboard"
     elif "Input" in user_agent:
         return "Mobile app"
@@ -1571,6 +1571,5 @@ def parse_user_agent(user_agent: str) -> str:
         return "Synced from Media Sync"
     elif "Python-client" in user_agent:
         return "Mergin Maps Python Client"
-    else: # For uncommon user agent we display user agent as is
+    else:  # For uncommon user agent we display user agent as is
         return user_agent
-        
--- a/Mergin/utils.py
+++ b/Mergin/utils.py
@@ -399,14 +399,13 @@ def send_logs(username, logfile):
 
 def validate_mergin_url(url):
     """
-    Validation of mergin URL by pinging. Checks if URL points at compatible Mergin server.
+    Initiates connection to the provided server URL to check if the server is accessible
     :param url: String Mergin Maps URL to ping.
     :return: String error message as result of validation. If None, URL is valid.
     """
     try:
-        mc = MerginClient(url, proxy_config=get_qgis_proxy_config(url))
-        if not mc.is_server_compatible():
-            return "Incompatible Mergin Maps server"
+        MerginClient(url, proxy_config=get_qgis_proxy_config(url))
+
     # Valid but not Mergin URl
     except ClientError:
         return "Invalid Mergin Maps URL"
@@ -1223,7 +1222,7 @@ def test_server_connection(url, username, password):
     result = True, "<font color=green> OK </font>"
     proxy_config = get_qgis_proxy_config(url)
     try:
-        mc = MerginClient(url, None, username, password, get_plugin_version(), proxy_config)
+        MerginClient(url, None, username, password, get_plugin_version(), proxy_config)
     except (LoginError, ClientError) as e:
         QgsApplication.messageLog().logMessage(f"Mergin Maps plugin: {str(e)}")
         result = False, f"<font color=red> Connection failed, {str(e)} </font>"
--- a/Mergin/utils.py
+++ b/Mergin/utils.py
@@ -1509,15 +1509,6 @@ def get_layer_by_path(path):
             return layer
 
 
-def check_mergin_subdirs(directory):
-    """Check if the directory has a Mergin Maps project subdir (.mergin)."""
-    for root, dirs, files in os.walk(directory):
-        for name in dirs:
-            if name == ".mergin":
-                return os.path.join(root, name)
-    return False
-
-
 def contextual_date(date_string):
     """Converts datetime string returned by the server into contextual duration string, e.g.
     'N hours/days/month ago'
--- a/Mergin/utils.py
+++ b/Mergin/utils.py
@@ -1527,8 +1527,8 @@ def contextual_date(date_string):
     now = datetime.now(timezone.utc)
     delta = now - dt
     if delta.days > 365:
-        years = now.year - dt.year - ((now.month, now.day) < (dt.month, dt.day))
-        return f"{years} {'years' if years > 1 else 'year'} ago"
+        # return the date value for version older than one year
+        return dt.strftime("%Y-%m-%d")
     elif delta.days > 31:
         months = int(delta.days // 30.436875)
         return f"{months} {'months' if months > 1 else 'month'} ago"
--- a/Mergin/utils.py
+++ b/Mergin/utils.py
@@ -1556,16 +1556,21 @@ def format_datetime(date_string):
 
 
 def parse_user_agent(user_agent: str) -> str:
-    user_agent = user_agent.split()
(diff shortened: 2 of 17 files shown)


ValentinBuira:
You can find the artificat for this PR : https://github.com/MerginMaps/qgis-plugin/actions/runs/11256676320

I'll update the link for future version

saberraz:
Another observation, it takes a while for plugin:

- to cache the older version
- to render the map (specifically the background map and not the changes)
image

^^ An example of caching and creating the map

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,588 @@
+from collections import deque
```
tomasMizera, line 1: Missing licence header

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,588 @@
...
+    same_dir,
+    unhandled_exception_message,
+    unsaved_project_check,
+    UnsavedChangesStrategy,
+    contextual_date,
+    is_versioned_file,
+    icon_path,
+    format_datetime,
+    parse_user_agent
+)
```
tomasMizera, line 48: Quite a few of these imports are not used in this file

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,588 @@
...
+    def exec(self):
+        try:
+            ws_id = self.mp.workspace_id()
+        except ClientError as e:
+            QMessageBox.warning(None, "Client Error", str(e))
+            return
+
+        # check if user has permissions
+        usage = self.mc.workspace_usage(ws_id)
+        if not usage["view_history"]["allowed"]:
```
tomasMizera, line 376: Mind that `/workspace/<ws_id>/usage` endpoint might not be available on all servers (EE, CE)

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,588 @@
...
+            QMessageBox.warning(None, "Permission Error", "The workspace does not allow to view project history.")
```
tomasMizera, line 377: How about something like

```suggestion
            QMessageBox.warning(None, "Upgrade required", "To view the project history, please upgrade your subscription plan.")
```

to let users know what to do to see the history? 
Optionally also with a button to navigate to the subscription page :)

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,588 @@
...
+
+    finished = pyqtSignal(list)
+
+    def __init__(self, mc: MerginClient, project_name, model: VersionsTableModel, is_sync=False):
+        super(VersionsFetcher, self).__init__()
+        self.mc = mc
+        self.project_name = project_name
+        self.model = model
+
+        self.is_sync = is_sync
```
tomasMizera, line 227: This `is_sync` seems unused

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,588 @@
...
+            height = max([self.map_canvas.minimumSizeHint().height(), self.attribute_table.minimumSizeHint().height()])
+            self.splitter.setSizes([height, height])
+
+    def fetch_from_server(self):
+
+        if self.fetcher and self.fetcher.isRunning():
+            # Only fetching when previous is finshed
+            return
+
+        self.fetcher = VersionsFetcher(self.mc, self.mp.project_full_name(), self.versionModel)
```
tomasMizera, line 413: If you want to pass in the `versionModel` to the fetcher, I'd expect it would add the versions to the model directly, if not, do we even need to pass it in?

tomasMizera (commented):
A bunch of comments from a quick view

Mergin/version_viewer_dialog.py
ValentinBuira: I added one, but there is a lot of missing header in the plugin, I'll do a separate PR to correct it

Mergin/version_viewer_dialog.py
ValentinBuira: Removed those in last version

Mergin/utils.py
```diff
@@ -1507,3 +1507,65 @@ def get_layer_by_path(path):
...
+
+def parse_user_agent(user_agent: str) -> str:
+    user_agent = user_agent.split()
+    # Does the python-api-client have a specific user_agent ?
+    if len(user_agent) == 4:  # Plugin
+        _, plugin_version, qgis_version, platform = user_agent
+        plugin_version = plugin_version.replace("Plugin/", "")
+        qgis_version = qgis_version.replace("QGIS/", "")
+        platform = platform.strip("()")
+        return f"MerginMaps plugin version {plugin_version} and QGIS version {qgis_version} on {platform}"
```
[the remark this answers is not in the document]
ValentinBuira: Yes that's a good compromise

Mergin/diff.py
```diff
@@ -446,3 +530,17 @@ def style_diff_layer(layer, schema_table):
         )
         r = QgsRuleBasedRenderer(root_rule)
         layer.setRenderer(r)
+
+
+def icon_for_layer(layer) -> QIcon:
```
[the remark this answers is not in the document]
ValentinBuira: Could be both I moved it in utils anyway

Mergin/utils.py
```diff
@@ -1507,3 +1507,65 @@ def get_layer_by_path(path):
...
+    delta = now - dt
+    if delta.days > 365:
+        years = now.year - dt.year - ((now.month, now.day) < (dt.month, dt.day))
+        return f"{years} {'years' if years > 1 else 'year'} ago"
+    elif delta.days > 31:
+        months = int(delta.days // 30.436875)
+        return f"{months} {'months' if months > 1 else 'month'} ago"
+    elif delta.days > 6:
+        weeks = int(delta.days // 7)
+        return f"{weeks} {'weeks' if weeks > 1 else 'week'} ago"
```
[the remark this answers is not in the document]
ValentinBuira: I tried to dispay as year-month-day but I don't have old project on my hands to test, if you can have a look @alexd0793 ?

Mergin/utils.py
```diff
@@ -1507,3 +1507,65 @@ def get_layer_by_path(path):
...
+def format_datetime(date_string):
+    """Formats datetime string returned by the server into human-readable format"""
+    dt = datetime.strptime(date_string, "%Y-%m-%dT%H:%M:%SZ")
+    return dt.strftime("%a, %d %b %Y %H:%M:%S GMT")
+
+
+def parse_user_agent(user_agent: str) -> str:
+    user_agent = user_agent.split()
+    # Does the python-api-client have a specific user_agent ?
+    if len(user_agent) == 4:  # Plugin
```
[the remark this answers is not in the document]
ValentinBuira: I changed this to encompass 97% on user agent

Mergin/version_viewer_dialog.py
ValentinBuira: remove in latest refactor

Mergin/version_viewer_dialog.py
ValentinBuira: Moved this into a simple try/except block

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,639 @@
...
+            height = max([self.map_canvas.minimumSizeHint().height(), self.attribute_table.minimumSizeHint().height()])
+            self.splitter_map_table.setSizes([height, height])
+
+    def fetch_from_server(self):
+
+        if self.fetcher and self.fetcher.isRunning():
+            # Only fetching when previous is finshed
+            return
+        else:
+            self.fetcher.start()
```
uclaros, line 416: Is there any visual indication that we are fetching?

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,639 @@
...
+        super().exec()
+
+    def closeEvent(self, event):
+        self.save_splitters_state()
+        QDialog.closeEvent(self, event)
+
+    def save_splitters_state(self):
+        settings = QSettings()
+        settings.setValue("Mergin/VersionViewerSplitterSize", self.splitter_map_table.saveState())
+        settings.setValue("Mergin/VersionViewerSplitterVericalSize", self.splitter_vertical.saveState())
```
uclaros, line 388:
```suggestion
        settings.setValue("Mergin/versionViewerSplitterSize", self.splitter_map_table.saveState())
        settings.setValue("Mergin/versionViewerSplitterVericalSize", self.splitter_vertical.saveState())
```
Settings entries are lowerCamelCase

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,639 @@
...
+
+    The __init__ method follow this pattern after varaible initiatlization
+    the methods of the class also follow this pattern
+    """
+
+    def __init__(self, mc, parent=None):
+
+        QDialog.__init__(self, parent)
+        self.ui = uic.loadUi(ui_file, self)
+
```
uclaros, line 264: `QgsGui.instance().enableAutoGeometryRestore(self)` should allow for QGIS to resize the dialog to it's last size/pos

Mergin/utils.py
```diff
@@ -1507,3 +1507,84 @@ def get_layer_by_path(path):
...
             return layer
+
+
+def check_mergin_subdirs(directory):
+    """Check if the directory has a Mergin Maps project subdir (.mergin)."""
+    for root, dirs, files in os.walk(directory):
+        for name in dirs:
+            if name == ".mergin":
+                return os.path.join(root, name)
+    return False
```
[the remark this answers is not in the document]
ValentinBuira: Indeed thanks for the catch

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,639 @@
...
+from .mergin.utils import bytes_to_human_size, int_version
+
+from .mergin import MerginClient
+from .diff import make_version_changes_layers
+
+
+ui_file = os.path.join(os.path.dirname(os.path.realpath(__file__)), "ui", "ui_versions_viewer.ui")
+
+
+class VersionsTableModel(QAbstractTableModel):
```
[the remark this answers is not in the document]
ValentinBuira: Actually it's the other way around, we do use a `QTreeView` but a `QAbstractTableModel` as a model, because right now  the data are flat and `QAbstractTableModel` provides convenient already implemented methods to work with tabular data.

If we want to  expand to tree we would have to change in the feature but have to write more boilre plate code

Mergin/version_viewer_dialog.py
ValentinBuira: Not at the moment, I think the cleanest way to do so would be to add a temporary items at the end of the version table Saying "Fetching more versions ⏳"

Mergin/version_viewer_dialog.py
ValentinBuira: Added in latest commits as well

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,639 @@
...
+
+    def current_version_changed(self, current_index, previous_index):
+        # Update the ui when the selected version change
+        item = self.versionModel.item_from_index(current_index)
+        version_name = item["name"]
+        version = int_version(item["name"])
+
+        self.setWindowTitle(f"Changes Viewer | {version_name}")
+
+        self.version_details = self.mc.project_version_info(self.mp.project_id(), version_name)
```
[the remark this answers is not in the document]
ValentinBuira: Yes the key are guaranteed
And if the user doesn't have access to internet when openning the project the icon are grayed out like that: 
Screenshot From 2024-12-12 12-27-16

**Howewer if the user disconnect from internet in the middle of the use of the plugin it would have failed with an error, I added an extra try/catch to prevent it**

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,639 @@
...
+        if self.fetcher and self.fetcher.isRunning():
+            # Only fetching when previous is finshed
+            return
+        else:
+            self.fetcher.start()
+
+    def on_scrollbar_changed(self, value):
+
+        if self.ui.history_treeview.verticalScrollBar().maximum() <= value:
+            self.fetch_from_server()
```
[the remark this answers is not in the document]
ValentinBuira: Yes but it's very unlikely to happen as we fetch 100 versions at once

Mergin/utils.py
```diff
@@ -1507,3 +1507,84 @@ def get_layer_by_path(path):
...
+
+        return f"{hours} {'hours' if hours > 1 else 'hour'} ago"
+
+    return f"{delta.days} {'days' if delta.days > 1 else 'day'} ago"
+
+
+def format_datetime(date_string):
+    """Formats datetime string returned by the server into human-readable format"""
+    dt = datetime.strptime(date_string, "%Y-%m-%dT%H:%M:%SZ")
+    return dt.strftime("%a, %d %b %Y %H:%M:%S GMT")
```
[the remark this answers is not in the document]
ValentinBuira: yes ideally `UTC` would be more appropriate however after testing the server store date as `GMT` if we use `UTC` the time is shifted by an hour

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,639 @@
...
+        self.mp = MerginProject(self.project_path)
+
+        self.set_splitters_state()
+
+        self.versionModel = VersionsTableModel()
+        self.history_treeview.setModel(self.versionModel)
+        self.history_treeview.verticalScrollBar().valueChanged.connect(self.on_scrollbar_changed)
+
+        self.selectionModel: QItemSelectionModel = self.history_treeview.selectionModel()
+        self.selectionModel.currentChanged.connect(self.current_version_changed)
```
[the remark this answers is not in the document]
ValentinBuira: Will functionally `QTreeView::currentChanged` would require subclassing QTreeView while `QItemSelectionModel::currentChanged` works with signal and slot

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,639 @@
...
+            self.diff_downloader.finished.connect(lambda msg: self.show_version_changes(version))
+            self.diff_downloader.start()
+        else:
+            self.show_version_changes(version)
+
+    def populate_details(self):
+        self.edit_project_size.setText(bytes_to_human_size(self.version_details["project_size"]))
+        self.edit_created.setText(format_datetime(self.version_details["created"]))
+        self.edit_user_agent.setText(parse_user_agent(self.version_details["user_agent"]))
+        self.edit_user_agent.setToolTip(self.version_details["user_agent"])
```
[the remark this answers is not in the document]
ValentinBuira: They are marked as readOnly in the UI file, and the cursor don't change when hovering them but they do look  similar on face value :-/ we could maybe change the style of them ?

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,639 @@
...
+
+class VersionsTableModel(QAbstractTableModel):
+    VERSION = Qt.UserRole + 1
+    VERSION_NAME = Qt.UserRole + 2
+
+    def __init__(self, parent=None):
+        super().__init__(parent)
+
+        # Keep ordered
+        self.versions = deque()
```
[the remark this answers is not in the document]
ValentinBuira: It was because in one of the first iteration the history viewer was in a dock and not in a separate window, therefore user could synchronize project while viewing the history. So we kept tracks of versions in a queue.

Hower that's not need anymore and I revert to a simple list

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,660 @@
...
+
+    def selected_version_changed(self, current_index, previous_index):
+        # Update the ui when the selected version change
+        item = self.versionModel.item_from_index(current_index)
+        version_name = item["name"]
+        version = int_version(item["name"])
+
+        self.setWindowTitle(f"Changes Viewer | {version_name}")
+
+        self.version_details = self.mc.project_version_info(self.mp.project_id(), version_name)
```
uclaros, line 452: This is also done on the main thread slowing down the ui response when the user clicks.

Could this be run in a separate thread?
Could we at least cache these responses? (are they needed when there is already a `.cache` folder?)

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,660 @@
...
+        if self.current_diff.isSpatial() == False:
+            self.map_canvas.setEnabled(False)
+            self.save_splitters_state()
+            self.splitter_map_table.setSizes([0, 1])
+        else:
+            self.map_canvas.setEnabled(True)
+            self.set_splitters_state()
+
+        self.update_canvas_layers(layers)
+        self.update_canvas_extend(layers)
```
uclaros, line 533: I'd remove these two methods for simplicity and add their code in `update_canvas()` instead.
It will also save us a refresh.

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,660 @@
...
+        if layers:
+            extent = layers[0].extent()
+            d = min(extent.width(), extent.height())
+            if d == 0:
+                d = 1
+            extent = extent.buffered(d * 0.07)
+
+            extent = self.map_canvas.mapSettings().layerExtentToOutputExtent(layers[0], extent)
+
+            self.map_canvas.setExtent(extent)
```
uclaros, line 551: Not sure what you're trying to do here...
Is it:
```suggestion
            extent = self.map_canvas.mapSettings().layerExtentToOutputExtent(layers[0], extent)
            extent.scale(1.07)
            self.map_canvas.setExtent(extent)
```
?

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,660 @@
...
+            self.toolbar.addAction(self.zoom_selected_action)
+
+            btn_add_changes = QPushButton("Add to project")
+            btn_add_changes.setIcon(QgsApplication.getThemeIcon("/mActionAdd.svg"))
+            menu = QMenu()
+            add_current_action = menu.addAction(
+                QIcon(icon_path("file-plus.svg")), "Add current changes layer to project"
+            )
+            add_current_action.triggered.connect(self.add_current_to_project)
+            add_all_action = menu.addAction(QIcon(icon_path("folder-plus.svg")), "Add all changes layers to project")
```
uclaros, line 343: This is not super clear what these two action will do, and the fact that once clicked they will no longer do anything until a different version is selected does not help.

There's room for improvement here:
- Add tooltips that will say that temporary scratch layers will be created
- Maybe disable them once clicked?

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,660 @@
...
+            # transform extent
+            layerExtent = self.map_canvas.mapSettings().layerExtentToOutputExtent(self.current_diff, layerExtent)
+
+            self.map_canvas.setExtent(layerExtent)
+            self.map_canvas.refresh()
+
+    def zoom_selected(self):
+        if self.current_diff:
+            self.map_canvas.zoomToSelected([self.current_diff])
+            self.map_canvas.refresh()
```
uclaros, line 654: A little pedantic, but neither the list nor the refresh should be needed here.
```suggestion
            self.map_canvas.zoomToSelected(self.current_diff)
```

Mergin/diff.py
```diff
@@ -311,6 +330,80 @@ def make_local_changes_layer(mp, layer):
...
+            db_conn = sqlite3.connect(gpkg_file)
+
+            features = diff_table_to_features(diff[table_name], db_schema[table_name], fields, cols_to_fields, db_conn)
+
+            # create diff layer
+            if geom_type is None:
+                continue
+
+            uri = f"{geom_type}?crs=epsg:{geom_crs}" if geom_crs else geom_type
+            vl = QgsVectorLayer(uri, table_name, "memory")
```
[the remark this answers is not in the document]
ValentinBuira: Not sure what you mean ? The current version is displayed everywhere (on the left panel, on the title etc...)

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,660 @@
...
+        elif role == VersionsTableModel.VERSION:
+            return int_version(self.versions[idx]["name"])
+        elif role == VersionsTableModel.VERSION_NAME:
+            return self.versions[idx]["name"]
+        else:
+            return None
+
+    def insertRows(self, row, count, parent=QModelIndex()):
+        self.beginInsertRows(parent, row, row + count - 1)
+        self.endInsertRows()
```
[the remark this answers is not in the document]
ValentinBuira: I was under the impression I had to re implement insertRows but actually I don't I moved begin/endInsertRows() in the appropriate function

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,660 @@
...
+
+    def clear(self):
+        self.beginResetModel()
+        self.versions.clear()
+        self.endResetModel()
+
+    def add_versions(self, versions):
```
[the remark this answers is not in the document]
ValentinBuira: I prefer append as it matches python and I renamed it however prepend_versions was not used anymore and I deleted it

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,660 @@
...
+        self.map_canvas.setLayers([])
+        self.attribute_table.clearSelection()
+
+        self.current_diff = self.diff_layers[index]
+
+        self.layer_cache = QgsVectorLayerCache(self.current_diff, 1000)
+        self.layer_cache.setCacheGeometry(False)
+
+        self.table_model = QgsAttributeTableModel(self.layer_cache)
+        self.table_model.setRequest(QgsFeatureRequest().setFlags(QgsFeatureRequest.NoGeometry).setLimit(100))
```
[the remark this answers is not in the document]
ValentinBuira: Not sure the limitation also exists in the diff viewer dialog, I'll ask why

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,660 @@
...
+        return items
+
+    def _table_summary_items(self, summary):
+        return [QStandardItem("{}: {}".format(k, summary[k])) for k in summary if k != "table"]
+
+    def toggle_project_layers(self, checked):
+        if checked:
+            self.toggle_layers_button.setText("Hide background layers")
+        else:
+            self.toggle_layers_button.setText("Show background layers")
```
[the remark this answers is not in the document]
ValentinBuira: Unfortunately having action and a button is the only way I found to display both the icon AND the text beside it.

However there was a mismatch between the text and the state on the first click I fixed it

Mergin/version_viewer_dialog.py
ValentinBuira: I think the intend is the same is both code, I'm not sure why I some point we defaulted to 1 ?

Mergin/version_viewer_dialog.py
ValentinBuira: This matches the behavior in the diff  viewer dialog, updating the tooltip if fine but I would consider anything else as nice to have for futures improvements

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,660 @@
...
+            self.has_selected_latest = False
+
+            try:
+                self.fetcher = VersionsFetcher(self.mc, self.mp.project_full_name(), self.versionModel)
+                self.diff_downloader = None
+
+                self.fetcher.fetch_another_page()
+            except ClientError as e:
+                self.failed_to_fetch = True
+                return
```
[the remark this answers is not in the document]
ValentinBuira: Indeed I replaced it  with another function that either check if the thread is running otherwise start it

Mergin/diff.py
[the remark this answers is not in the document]
uclaros: > _version number_ to the layer name

I mean the name of the temporary scratch layer added, so that:

- User selects version 42
- Clicks add layer
- Layer gets added as `table_name (v42)`
- user selects version 41
- Clicks add layer
- Layer gets added as `table_name (v41)`
- Project now has two memory layers that the user can distinguish between, `table_name (v42)` and `table_name (v41)`

Mergin/version_viewer_dialog.py
tomasMizera: We agreed to add the footer

Mergin/diff.py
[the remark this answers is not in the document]
ValentinBuira: Added in this commit : https://github.com/MerginMaps/qgis-plugin/pull/630/commits/d20edd7285e5f9ba96205e62a2bfe5cba6ab5481

Mergin/version_viewer_dialog.py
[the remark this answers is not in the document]
ValentinBuira: In the end we settle on only display "Background" layer and let the checked UI to convey the meaning

Mergin/version_viewer_dialog.py
ValentinBuira: Added as part of https://github.com/MerginMaps/qgis-plugin/pull/630/commits/fefe3a61b165c0535125f40a43ae5e026f296e56 it addes quite a bit of overhead as we temporary increase `rowCount `on fetch

Mergin/version_viewer_dialog.py
```diff
@@ -0,0 +1,660 @@
...
+    def __init__(self, mc: MerginClient, project_path, model: VersionsTableModel):
+        super(VersionsFetcher, self).__init__()
+        self.mc = mc
+        self.project_path = project_path
+        self.model = model
+
+        self.current_page = 1
+        self.per_page = 50
+
+        version_count = self.mc.project_versions_count(self.project_path)
```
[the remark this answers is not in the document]
ValentinBuira: Yes but we need these info right away, we agreed with @tomasMizera  to let it as is for now as it would require more refactor to make this async

Mergin/version_viewer_dialog.py
ValentinBuira: Same as above

Mergin/version_viewer_dialog.py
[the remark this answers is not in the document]
ValentinBuira: implemented with a callback at the end of fetch to fetch again if scrollbar are not visible see https://github.com/MerginMaps/qgis-plugin/pull/630/commits/264e1614a418684fe497f5989b2567857944ea6f

tomasMizera:
Let's merge this now and continue fixing from `master` 🎉

tomasMizera (approved):
image

PR #632, opened 2024-10-10
Fix 617: Reload project following sync of updated .qgz file on windows

### LINKED ISSUE #617: Reload project following download of updated .qgz file
### Summary

As a Mergin Maps user, I would like the plugin to reload the QGIS project when a new version of the .qgz file has been downloaded, so that I see changes that have been made by other users.

### Description

Consider this scenario:

+ User 1 (Writer permissions) and User 2 (Editor permissions) both synchronise and have identical projects.
+ User 1 adds a geotiff with an aerial photo to the project folder and adds it as a layer in the project so it displays on the map.
+ User 1 saves the project and syncs the files.
+ User 2 presses "syncronise".  The dialog says that there are changes to the .qgz file.  They press OK.

What happens:

+ The geotiff is downloaded to User 2's laptop and the plugin reports successful synchronisation.
+ The aerial photo does not appear on the map.
+ If User 2 closes and re-opens the project, then the aerial photo appears.

What should happen:

+ The project is reloaded so that the aerial photo appears as part of the synchronisation process

### Related issues

https://github.com/MerginMaps/qgis-plugin/issues/592

Plugin version: 2024.2

ValentinBuira:
The qgis project was not reloading after a sync on windows when there was only change to the .qgz file

This happened because ` QgsProject.instance().fileName()` &` QgsProject.instance().absolutePath() `returned unix-style path even on windows. This behavior mismatch with the return of e.g `os.path.join` in python. 

This PR fix the is by adding an extra os.path.normpath

Fix #617

### DIFF
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -715,13 +715,6 @@ class MerginLocalProjectItem(QgsDirectoryItem):
             return
         self.project_manager.project_status(self.path)
 
-    def _reload_project(self):
-        """This will forcefully reload the QGIS project because the project (or its data) may have changed"""
-        qgis_proj_filename = os.path.normpath(QgsProject.instance().fileName())
-        qgis_files = find_qgis_files(self.path)
-        if qgis_proj_filename in qgis_files:
-            iface.addProject(QgsProject.instance().fileName())
-
     def remove_local_project(self):
         if not self.path:
             return
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -717,8 +717,9 @@ class MerginLocalProjectItem(QgsDirectoryItem):
 
     def _reload_project(self):
         """This will forcefully reload the QGIS project because the project (or its data) may have changed"""
+        qgis_proj_filename = os.path.normpath(QgsProject.instance().fileName())
         qgis_files = find_qgis_files(self.path)
-        if QgsProject.instance().fileName() in qgis_files:
+        if qgis_proj_filename in qgis_files:
             iface.addProject(QgsProject.instance().fileName())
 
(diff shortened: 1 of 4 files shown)


Mergin/plugin.py
```diff
@@ -717,8 +717,9 @@ def sync_project(self):
 
     def _reload_project(self):
         """This will forcefully reload the QGIS project because the project (or its data) may have changed"""
+        qgis_proj_filename = os.path.normpath(QgsProject.instance().fileName())
         qgis_files = find_qgis_files(self.path)
-        if QgsProject.instance().fileName() in qgis_files:
+        if qgis_proj_filename in qgis_files:
             iface.addProject(QgsProject.instance().fileName())
```
uclaros, line 723: This whole `_reload_project()` method seems unused. Could we remove it or am I missing something?

uclaros (commented):
Looks good, one minor suggestion!

Also, it would be nice if at some point we standardized the way paths are handled though, as now there are combinations of QT style (forward slashes, used by qgis api), os.path and pathlib.Path

ValentinBuira:
> Also, it would be nice if at some point we standardized the way paths are handled though, as now there are combinations of QT style (forward slashes, used by qgis api), os.path and pathlib.Path

Agreed I think pathlib is the good candidate and the more pleasant to work with

Mergin/plugin.py
ValentinBuira: Indeed it is unused I removed it in the last commit

ValentinBuira: Thanks for the catch

volcan01010:
Thank you for fixing this so quickly.

PR #637, opened 2024-10-18
Rename deprecated enum and method in qt5 to qt6

ValentinBuira:
This PR rename deprecated enum and method in qt5 and that will be removed in qt6. This should make the plugin mostly compatible when QGIS finally does switch to a newer version of qt.

See : https://www.pythonguis.com/faq/pyqt5-vs-pyqt6/

Ideally from now on we should  write code compatible with qt6

### DIFF
--- a/Mergin/clone_project_dialog.py
+++ b/Mergin/clone_project_dialog.py
@@ -20,7 +20,7 @@ class CloneProjectDialog(QDialog):
         """
         QDialog.__init__(self)
         self.ui = uic.loadUi(ui_file, self)
-        self.ui.buttonBox.button(QDialogButtonBox.Ok).setEnabled(False)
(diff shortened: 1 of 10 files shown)


uclaros:
Looks good.
Did you use https://github.com/qgis/QGIS/blob/master/scripts/pyqt5_to_pyqt6/pyqt5_to_pyqt6.py for that?
If not, maybe you could run that too to see if something else pops up!

ValentinBuira:
Thanks for the review, I didn't use a script for this PR

I tried the pyqt5_to_pyqt6.py script you mentioned and it brings a LOT of other change I missed. And some change I don't understand the point.

I'll merge this one for now and have  a second looks with the automated script

PR #640, opened 2024-11-11
New Monthly Contributors Error Dialog

### LINKED ISSUE #643: Show dialog when active contributors per month limit is hit [plugin + py-client]
Let's show a nice dialogue for QGIS Plugin users, similar to 
https://www.figma.com/design/Zl8irhQtrHrxBVcJyLPedL/Dashboard-2.0?node-id=1604-8568&t=rrDFDGoGMb7lsc0X-4

Plus, let's make sure that the error is understandable on py-client and db-sync too

VitorVieiraZ:
Displays `Monthly Contributos Error Dialog` when creating a project or syncing in case of failure.

Resolves #643, Python-api-client update: https://github.com/MerginMaps/python-api-client/pull/221

### DIFF
--- a/Mergin/configuration_dialog.py
+++ b/Mergin/configuration_dialog.py
@@ -14,7 +14,7 @@ except ImportError:
     this_dir = os.path.dirname(os.path.realpath(__file__))
     path = os.path.join(this_dir, "mergin_client.whl")
     sys.path.append(path)
-    from mergin.client import MerginClient, ClientError, LoginError
+    from .mergin.client import MerginClient, ClientError, LoginError
 
 from .utils import (
     get_mergin_auth,
--- a/Mergin/configuration_dialog.py
+++ b/Mergin/configuration_dialog.py
@@ -14,7 +14,7 @@ except ImportError:
     this_dir = os.path.dirname(os.path.realpath(__file__))
     path = os.path.join(this_dir, "mergin_client.whl")
     sys.path.append(path)
-    from .mergin.client import MerginClient, ClientError, LoginError
+    from mergin.client import MerginClient, ClientError, LoginError
 
 from .utils import (
     get_mergin_auth,
--- a/Mergin/monthly_contributors_error_dialog.py
+++ b/Mergin/monthly_contributors_error_dialog.py
@@ -25,7 +25,9 @@ class MonthlyContributorsErrorDialog(QDialog):
         upgrade_button.setText("Upgrade")
 
         quota = self.server_response.get("contributors_quota", "#NA")
-        quota_text = f"You've reached the maximum number of active monthly contributors ({quota}) for your current subscription."
+        quota_text = (
+            f"You've reached the maximum number of active monthly contributors ({quota}) for your current subscription."
+        )
         self.label.setText(quota_text)
 
     def open_upgrade_link(self):
--- a/Mergin/monthly_contributors_error_dialog.py
+++ b/Mergin/monthly_contributors_error_dialog.py
@@ -8,6 +8,7 @@ from .help import MerginHelp
 
 ui_file = os.path.join(os.path.dirname(os.path.realpath(__file__)), "ui", "ui_monthly_contributors_error_dialog.ui")
 
+
 class MonthlyContributorsErrorDialog(QDialog):
     def __init__(self, e, parent=None):
         QDialog.__init__(self, parent)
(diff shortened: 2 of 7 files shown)


Mergin/configuration_dialog.py
```diff
@@ -14,7 +14,7 @@
     this_dir = os.path.dirname(os.path.realpath(__file__))
     path = os.path.join(this_dir, "mergin_client.whl")
     sys.path.append(path)
-    from mergin.client import MerginClient, ClientError, LoginError
+    from .mergin.client import MerginClient, ClientError, LoginError
```
ValentinBuira, line 17: This should work without the `dot` ?

Mergin/monthly_contributors_error_dialog.py
```diff
@@ -0,0 +1,40 @@
...
+
+class MonthlyContributorsErrorDialog(QDialog):
+    def __init__(self, e, parent=None):
+        QDialog.__init__(self, parent)
+        self.ui = uic.loadUi(ui_file, self)
+
+        self.server_response = e.server_response
+        self.set_dialog_style()
+
+        self.cancel_btn.clicked.connect(self.reject)
```
ValentinBuira, line 18: Just for your information when you want only standard action there is a [QDialogButtonBox](https://doc.qt.io/qt-5/qdialogbuttonbox.html) to layout out standard button 

But you have less fine grained support to lay them as you want

ValentinBuira (changes requested):
Just few things I am not quite sure to understand otherwise look good to me

I added few general tips feel free to ignore them and mark them as solved

ValentinBuira (approved):

Mergin/projects_manager.py
```diff
@@ -95,6 +96,10 @@ def create_project(self, project_name, project_dir, is_public, namespace):
             # User friendly error messages
             if e.http_error == 409:
                 msg = f'Project named "{project_name}" already exists in the workspace "{namespace}".\nPlease try renaming the project.'
+            elif e.http_error == 422:
```
tomasMizera, line 99: We can not rely only on the http error response, there could be multiple options for this code. Instead, we need to rely on our defined error codes (see https://github.com/MerginMaps/server-private/blob/fda7fc34bdbad91827391ebf4f525f4abd0c8e5b/server/src/service/errors.py#L70). It should be something like:

```suggestion
            elif e.server_code == ErrorCode.MonthlyContributorsLimitHit.value:
```

Mergin/projects_manager.py
```diff
@@ -375,6 +379,10 @@ def sync_project(self, project_dir, project_name=None):
                 if dlg.exception.http_error == 400 and "Another process" in dlg.exception.detail:
                     # To note we check for a string since error in flask doesn't return server error code
                     msg = "Somebody else is syncing, please try again later"
+                elif dlg.exception.http_error == 422:
```
tomasMizera, line 382: Same comment as the previous one :)

Mergin/ui/ui_monthly_contributors_error_dialog.ui
```diff
@@ -0,0 +1,135 @@
...
+     </property>
+     <property name="font">
+      <font>
+       <pointsize>16</pointsize>
+       <weight>75</weight>
+       <bold>true</bold>
+      </font>
+     </property>
+     <property name="text">
+      <string>You've reached the maximum number of active monthly contributors for your current subscription.</string>
```
tomasMizera, line 84: How about adding the number to the title instead of having the progress bar? The thing is that the progress bar would always be full, so it does not indicate progress of any kind :)

```suggestion
      <string>You've reached the maximum number of active monthly contributors (quota) for your current subscription.</string>
```

tomasMizera (changes requested):

Mergin/monthly_contributors_error_dialog.py
```diff
@@ -0,0 +1,35 @@
...
+        self.buttonBox.rejected.connect(self.reject)
+
+    def set_dialog_style(self):
+        upgrade_button = self.buttonBox.button(QDialogButtonBox.Ok)
+        upgrade_button.setText("Upgrade")
+
+        quota = self.server_response.get("contributors_quota", "#NA")
+        quota_text = (
+            f"You've reached the maximum number of active monthly contributors ({quota}) for your current subscription."
+        )
```
tomasMizera, line 30: Just some nit picking - this might seem more friendly for users in case we have issues reading the quota

```suggestion
        quota = self.server_response.get("contributors_quota", None)
        quota_text = (
            f"You've reached the maximum number of active monthly contributors {'(' + quota + ')' if quota else ''} for your current subscription."
        )
```

tomasMizera (approved):
Looking good on the plugin side, only some tiny comment. Feel free to squash and merge after addressing it!

tomasMizera:
Let's just bear in mind that we need a new Python-api-client version when releasing this one!

tomasMizera:
This work is not needed now with the recent changes

PR #641, opened 2024-11-12
Script to format python files

VitorVieiraZ:
Format Python files in `Mergin` folder using Black code formatter.

Setup:
1. Install Black: `pip install black`
2. Make script executable: `chmod +x format_py.bash`
3. Run script `./format_py.bash`

### DIFF
--- a/docs/dev-docs.md
+++ b/docs/dev-docs.md
@@ -31,3 +31,12 @@ Now link the plugin to your QGIS profile python:
 
 Plugin packages are built as GitHub actions for every commit.
 When releasing, make sure to check if [run-test.yml](../.github/workflows/run-test.yml) is using the latest QGIS release tag for auto-tests.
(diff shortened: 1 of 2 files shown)


tomasMizera:
Just a note, there is some formatting in CI: https://github.com/MerginMaps/qgis-plugin/blob/master/.github/workflows/code_style.yml

It would also be good to state this in developer docs - how to format the code and what you need to install

VitorVieiraZ:
> Just a note, there is some formatting in CI: https://github.com/MerginMaps/qgis-plugin/blob/master/.github/workflows/code_style.yml
> 
> It would also be good to state this in developer docs - how to format the code and what you need to install

Thanks for the note. The script is set up to handle formatting to pass the CI check.
I have updated developer docs on how to use the script.

PR #648, opened 2024-11-29
Remove deprecated server validation use

tomasMizera:
As a follow-up of https://github.com/MerginMaps/python-api-client/pull/223. Removes outdated validation of a server URL.

Requires a new py-api release.

### DIFF
--- a/Mergin/utils.py
+++ b/Mergin/utils.py
@@ -399,14 +399,13 @@ def send_logs(username, logfile):
 
 def validate_mergin_url(url):
     """
-    Validation of mergin URL by pinging. Checks if URL points at compatible Mergin server.
+    Initiates connection to the provided server URL to check if the server is accessible
     :param url: String Mergin Maps URL to ping.
     :return: String error message as result of validation. If None, URL is valid.
     """
     try:
-        mc = MerginClient(url, proxy_config=get_qgis_proxy_config(url))
-        if not mc.is_server_compatible():
-            return "Incompatible Mergin Maps server"
+        _ = MerginClient(url, proxy_config=get_qgis_proxy_config(url))
+
     # Valid but not Mergin URl
     except ClientError:
         return "Invalid Mergin Maps URL"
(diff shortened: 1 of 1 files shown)


jsjtadie:
In Python, if you want to remove the use of deprecated server authentication, it depends on the specific library or framework you're using. For example, in some cases related to HTTP requests, you might need to update the way you handle authentication methods.
 
Let's say you were using an older version of  urllib  and had some deprecated server auth code like this:
 
import urllib.request

# Old deprecated way (for example's sake, might not be actual deprecated code like this)
auth_handler = urllib.request.HTTPBasicAuthHandler()
opener = urllib.request.build_opener(auth_handler)
urllib.request.install_opener(opener)
 
 
You'd need to switch to a more modern and recommended approach. For instance, using  requests  library which has a cleaner and up-to-date way to handle authentication:
 
import requests

# Using requests for authentication
response = requests.get('https://example.com', auth=('username', 'password'))

Mergin/utils.py
```diff
@@ -399,14 +399,13 @@ def send_logs(username, logfile):
...
-    Validation of mergin URL by pinging. Checks if URL points at compatible Mergin server.
+    Initiates connection to the provided server URL to check if the server is accessible
     :param url: String Mergin Maps URL to ping.
     :return: String error message as result of validation. If None, URL is valid.
     """
     try:
-        mc = MerginClient(url, proxy_config=get_qgis_proxy_config(url))
-        if not mc.is_server_compatible():
-            return "Incompatible Mergin Maps server"
+        _ = MerginClient(url, proxy_config=get_qgis_proxy_config(url))
```
ValentinBuira, line 407: The python convention is to use `_` as a throwaway when we absolutely need place holder variable (e.g unpack variable), here I think we can just remove it altogether

```suggestion
        MerginClient(url, proxy_config=get_qgis_proxy_config(url))
```

Mergin/utils.py
```diff
@@ -1223,7 +1222,7 @@ def test_server_connection(url, username, password):
     result = True, "<font color=green> OK </font>"
     proxy_config = get_qgis_proxy_config(url)
     try:
-        mc = MerginClient(url, None, username, password, get_plugin_version(), proxy_config)
+        _ = MerginClient(url, None, username, password, get_plugin_version(), proxy_config)
```
ValentinBuira, line 1225: same as above

tomasMizera:
Required dependency was merged, we can merge this one then too. Let's keep in mind that we first need to release py-client

ValentinBuira (approved):
Looks good to me ! Thanks for the changes

PR #650, opened 2024-12-03
Cleanup useless function and error handling 

ValentinBuira:
This PR contain two cleanup

* Remove useless "post_login" and "mc.user_service()" since user_service was deprecated and return http 400 the `post_login`  function didn't do anything
see https://github.com/MerginMaps/server-private/blob/7edf64b43b04bf017c8c555f913f6d8c3b0d7441/server/src/service/api.yaml#L42
* Handle new ClientError from `mc.workspace_service`

Rely on https://github.com/MerginMaps/python-api-client/pull/224

### DIFF
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -158,8 +158,6 @@ class MerginPlugin:
 
             self.enable_toolbar_actions()
 
-        self.post_login()
-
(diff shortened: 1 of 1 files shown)


tomasMizera (approved):
LGTM

PR #652, opened 2025-01-10
Bump upload-artifact and download-artifact action to v4

ValentinBuira:
Bump upload-artifact and download-artifact action to v4.

Some changes were required in the PR because in  the V4, artifact are immutable once created. You can't incrementally add file to an artifact anymore, we have to use the `actions/upload-artifact/merge@v4` see https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#merging-multiple-artifacts


tomasMizera (approved):
I compared the output between `v3` and `v4` and it looks the same 👌🏻  Thanks for adjusting this @ValentinBuira

PR #655, opened 2025-01-13
make plugin compatibile with Qt6

alexbruy:
Update plugin to be usable with Qt6 builds of QGIS.

I haven't checked every possible use-case, so there still might be some minor issues.


ValentinBuira:
Thanks for the PR, did you use https://github.com/qgis/QGIS/blob/master/scripts/pyqt5_to_pyqt6/pyqt5_to_pyqt6.py for that, or another script for this ?

I would  prefer to wait for #630 (Project history) to  be merged and released before make officially the switch to be Qt 6 compatible

alexbruy:
I used a couple of different tools, including the script from QGIS.

> I would prefer to wait for https://github.com/MerginMaps/qgis-plugin/pull/630 (Project history) to be merged and released before make officially the switch to be Qt 6 compatible

Whatever is best for you. I just need to run plugin on Qt6, so I updated it for myself and thought it might be useful for others too, hence the PR.

Mergin/create_project_wizard.py
```diff
@@ -360,14 +362,14 @@ def node_shown(self, node):
 
     def flags(self, idx):
         if idx.column() == self.LAYER_COL:
-            return Qt.ItemIsEnabled
+            return Qt.ItemFlags.ItemIsEnabled
```
ValentinBuira, line 365: The Qt.ItemFlag enum should be used instead (without the s) otherwise it crashes when openning the creation wizard
```suggestion
            return Qt.ItemFlag.ItemIsEnabled
```

ValentinBuira (changes requested):

tomasMizera:
Let's merge this now, thanks guys for your work (cc @alexbruy)! We will conduct full regression tests on master before releasing this.

PR #660, opened 2025-02-11
Final adjustments for the history viewer 

ValentinBuira:
This PR does a bunch of quality of life improvement and fixes for the history viewer:
* Pre-select latest version
* Fix treeview not available when fetching
* GUI fix for macos
* Fix various out of sync issue with UI/thread
* Fix align left with TreeView decoration
* Update CI

### DIFF
--- a/Mergin/utils.py
+++ b/Mergin/utils.py
@@ -1580,7 +1580,7 @@ def icon_for_layer(layer) -> QIcon:
         return QgsApplication.getThemeIcon("/mIconTableLayer.svg")
 
 
-# Copy a layer and it's legend
+# Copy a layer and its legend
 def layer_copy(layer, style_name="mergin_diff"):
     lyr_clone = layer.materialize(QgsFeatureRequest().setFilterFids(layer.allFeatureIds()))
     lyr_clone.setName(layer.name())
--- a/Mergin/utils.py
+++ b/Mergin/utils.py
@@ -1581,10 +1581,8 @@ def icon_for_layer(layer) -> QIcon:
 
 
 # Copy a layer and it's legend
-def layer_copy(layer, style_name = 'mergin_diff'):
-    lyr_clone = layer.materialize(
-        QgsFeatureRequest().setFilterFids(layer.allFeatureIds())
-    )
+def layer_copy(layer, style_name="mergin_diff"):
+    lyr_clone = layer.materialize(QgsFeatureRequest().setFilterFids(layer.allFeatureIds()))
     lyr_clone.setName(layer.name())
     lyr_clone.setAttributeTableConfig(layer.attributeTableConfig())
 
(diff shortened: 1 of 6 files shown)


tomasMizera (approved):
Code is looking good, let me run it locally

Mergin/version_viewer_dialog.py
```diff
@@ -647,18 +668,15 @@ def diff_layer_changed(self, index: int):
 
     def add_current_to_project(self):
         if self.current_diff:
-            lyr_clone = QgsVectorLayer(
-                self.current_diff.source(),
-                self.current_diff.name() + f" ({self.version_details['name']})",
-                self.current_diff.providerType(),
-            )
+            lyr_clone = layer_copy(self.current_diff)
+            lyr_clone.setName(self.current_diff.name() + f" ({self.version_details['name']})")
```
uclaros, line 672: Why is the `layer_copy()` method needed?
Wouldn't this just work?
```suggestion
            lyr_clone = self.current_diff.clone()
            lyr_clone.setName(f"{lyr_clone.name()} ({self.version_details['name']})")
```

Mergin/utils.py
```diff
@@ -1578,3 +1578,23 @@ def icon_for_layer(layer) -> QIcon:
         return QgsApplication.getThemeIcon("/mIconGeometryCollectionLayer.svg")
     else:
         return QgsApplication.getThemeIcon("/mIconTableLayer.svg")
+
+
+# Copy a layer and it's legend
```
uclaros, line 1583:
```suggestion
# Copy a layer and its legend
```

Mergin/version_viewer_dialog.py
ValentinBuira: I wish 😢 Unfortunatly : 

> The [QgsVectorLayer](https://api.qgis.org/api/classQgsVectorLayer.html) class has a [clone()](https://qgis.org/api/classQgsVectorLayer.html#aa2fd518bbe4e03d7feb668f5774af4f3) method that allows you to clone the layer in a new layer, the problem is that if you modify the geometry in the cloned layer the original layer will be affected: the reason for this is that the data source is the same for the original layer and the cloned layer.

So if I clone the layer, and modify it, the cached data will be modified and end up being wrong in the history viewer

ValentinBuira: No you're right it's indeed memory layer, I tested layer.clone(), it partially work. However the conditional formatting is not cloned I guess I could use clone() and have a function to copy the style

uclaros: You can check how [`QgisApp::duplicateLayers()`](https://github.com/qgis/QGIS/blob/ac21a8d3a752f6f0e1440434cd1aea6c616d88e1/src/app/qgisapp.cpp#L11772) does it for reference.
`clone()`, then `exportNamedStyle()` - `importNamedStyle()`

ValentinBuira: Adapted the code from [QgisApp::duplicateLayers()](https://github.com/qgis/QGIS/blob/ac21a8d3a752f6f0e1440434cd1aea6c616d88e1/src/app/qgisapp.cpp#L11772)

PR #679, opened 2025-03-14
 Fix version history in older QGIS release 

ValentinBuira:
This PR fix the version history in older QGIS release notably:
* QgsTiledSceneLayer not available in QGIS < 3.34
* layerExtentToOutputExtent would fail in older version if extent was invalid (e.i without any geometry) so I added a check for this

This has the nice corollary to not display the map canvas when the layer has only null geometry instead of being to some random place in the ocean 

image

### DIFF
--- a/Mergin/version_viewer_dialog.py
+++ b/Mergin/version_viewer_dialog.py
@@ -20,9 +20,11 @@ from qgis.core import (
 try:
     from qgis.core import QgsTiledSceneLayer
 except ImportError:
+
     class QgsTiledSceneLayer:
         # Dummy class we only use this class to whitelist layers
         pass
+
 
 from qgis.gui import QgsAttributeTableFilterModel, QgsAttributeTableModel, QgsGui, QgsMapToolPan
 from qgis.PyQt import QtCore, uic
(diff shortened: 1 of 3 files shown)


Mergin/version_viewer_dialog.py
```diff
@@ -583,7 +594,11 @@ def update_canvas(self, layers, set_extent=True):
                 if d == 0:
                     d = 1
                 extent = extent.buffered(d * 0.07)
-                extent = self.map_canvas.mapSettings().layerExtentToOutputExtent(layers[0], extent)
+                extent = (
+                    extent
+                    if has_only_null_geometries(layers[0])
+                    else self.map_canvas.mapSettings().layerExtentToOutputExtent(layers[0], extent)
+                )
```
uclaros, line 601: Do we need to check all geometries in the layer (number of features could potentially be very big)?
Can't we just check if the extent is empty?

ValentinBuira: QGIS consider no extent to be `<QgsRectangle: -inf -inf, inf inf>` but inf is just the string representation of 2^1024

However I could do a `try/except ` in case layerExtentToOutputExtent

ValentinBuira: I settled to test the extent rectangle with `.isEmpty()` in the end, thanks !

uclaros (approved):

PR #684, opened 2025-04-01
Project home path issue

ldebek:


Mergin/validation.py
```diff
@@ -485,3 +493,5 @@ def warning_display_string(warning_id, url=None):
         return f"You don't have permission to change the configuration of this project. <a href='{url}'>Reset the configuration</a> to be able to sync data changes."
     elif warning_id == Warning.EDITOR_DIFFBASED_FILE_REMOVED:
         return f"You don't have permission to remove this layer. <a href='{url}'>Reset the layer</a> to be able to sync changes."
+    elif warning_id == Warning.PROJECT_HOME_PATH:
+        return "QGIS Project Home Path is specified. <a href='fix_project_home_path'>Fix the issue.</a>"
```
ValentinBuira, line 497: Silently reset the project home, we could warm the user just in case

```suggestion
        return "QGIS Project Home Path is specified. <a href='fix_project_home_path'>Quick fix the issue. (This will unset project home) </a>"
```

Mergin/utils.py
```diff
@@ -620,11 +620,19 @@ def set_qgis_project_relative_paths(qgis_project):
...
 
 
-def save_current_project(project_path, warn=False, relative_paths=True):
+def set_qgis_project_home_ignore(qgis_project):
+    """Check if given QGIS project have home path set up. If yes - remove it from the project."""
+    if qgis_project.presetHomePath():
+        qgis_project.setPresetHomePath("")
+
+
+def save_current_project(project_path, warn=False, relative_paths=True, ignore_project_home=True):
```
ValentinBuira, line 629: `ignore_project_home` is never used, same apply for the other arguments

Do you think you could clean the others as well ?

Mergin/utils.py
```diff
@@ -620,11 +620,19 @@ def set_qgis_project_relative_paths(qgis_project):
...
     """Save current QGIS project to project_path. Set the project to use relative paths if relative_paths is True."""
     cur_project = QgsProject.instance()
     if relative_paths:
         set_qgis_project_relative_paths(cur_project)
+    if ignore_project_home:
+        set_qgis_project_home_ignore(cur_project)
     cur_project.setFileName(project_path)
     write_success = cur_project.write()
     if not write_success and warn:
```
ValentinBuira, line 638: Will never happen because of the `warm` arguments

ValentinBuira (changes requested):
Hi @ldebek could you check the small cleanup in the `save_current_project` function ? But otherwise looks good to me

ldebek:
@ValentinBuira Requested changes applied.

PR #686, opened 2025-04-03
Initial prototype for SSO integration

wonder-sk:
- asks for email, then does auth and shows user info from MM endpoint using token from oauth2
- many todos left :upside_down_face:


tomasMizera:
I'll change the `base` into integration branch and merge the PR so that we can continue from here! :) Thanks @wonder-sk

PR #689, opened 2025-04-25
Drop support for out of date  server version (< 2023)

ValentinBuira:
This PR drop support for out of date server version (e.g version 2021.6.1) which was the first version of MM server 

Instead of checking thought the code base if the user are using an old server version. We prevent them to connect all together with a warning 

Screenshot From 2025-04-24 17-28-20


tomasMizera:
I believe there is more code to drop, for example https://github.com/MerginMaps/qgis-plugin/blob/68f85376de798c815642874a8460a880cc8a24bd/Mergin/plugin.py#L965-L966 and places like https://github.com/MerginMaps/qgis-plugin/blob/68f85376de798c815642874a8460a880cc8a24bd/Mergin/plugin.py#L1117

tomasMizera (approved):
It looks good just previewing the removed lines. @uclaros can I ask you to have a look at this one too? You introduced workspaces here so you might catch something extra.

After review, let's test everything all the related functionality and merge

Mergin/plugin.py
```diff
@@ -913,7 +901,7 @@ def update_client_and_manager(self, mc=None, manager=None, err=None):
     def updateName(self):
         name = self.base_name
         try:
-            if self.mc.server_type() != ServerType.OLD and self.plugin.current_workspace.get("name", None):
+            if self.plugin.current_workspace.get("name", None):
                 name = f"{self.base_name} [{self.plugin.current_workspace['name']}]"
         except AttributeError:
             # self.mc might not be set yet
```
uclaros, line 907: I believe `AttributeError` can no longer be raised. How about:
```suggestion
            name = f"{self.base_name} [{self.plugin.current_workspace['name']}]"
        except KeyError:
            # no current workspace
```
?

uclaros (commented):
This also needs some tweaking:
https://github.com/MerginMaps/qgis-plugin/blob/3046ad7abc1f45c01c2d00540e669f74383eed1b/Mergin/plugin.py#L402-L410

It's now empty just when there are no workspaces?
https://github.com/MerginMaps/qgis-plugin/blob/3046ad7abc1f45c01c2d00540e669f74383eed1b/Mergin/plugin.py#L91

ValentinBuira:
I removed even more code that was forgotten in the first pass, as you can see in the in the last three commits.

Sorry for the two part review

ValentinBuira:
I merged latest master back in and should be ready to be merged @uclaros @tomasMizera

JanCaha:
Looks good, but needs extensive testing...

PR #690, opened 2025-04-25
Add user friendly display expression for tracking layer

### LINKED ISSUE #681: Set display expression for tracking layer
Let's setup some nice display expression for tracking layer - we could use `username` and combination of dates.

It is currently empty :/

ValentinBuira:
When a tracking layer is enable we create a default display expression as well 

Looks like this in the mobile app

Fix #681


ValentinBuira:
This preview ? it automatically add a line break 

image

tomasMizera (approved):

PR #696, opened 2025-05-05
Fix unavailable mergin_project_* variables and remove outdated mergin_project_owner variable

ValentinBuira:
* Make (mm|mergin)_project_* variable available again in the plugin
* remove mergin_project_owner

### DIFF
--- a/Mergin/plugin.py
+++ b/Mergin/plugin.py
@@ -64,6 +64,7 @@ from .utils import (
     PROJS_PER_PAGE,
     remove_project_variables,
     same_dir,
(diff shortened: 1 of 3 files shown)


ValentinBuira:
Hi @tomasMizera, I checked in the mobile repository `mergin_project_server` is not available in the mobile app

In the plugin we read this variable from the local `QSettings` but it's a really weak approach because it only works if the the project was created in one specific QGIS installation.  I would advocate moving the remote server url in the metadata (mergin.json) so it could be read from both the plugin and the mobile app

PR #713, opened 2025-06-12
Move send of logs logic to mergin-client

JanCaha:
Simplify sending the logs by moving the logic to **mergin-client**


JanCaha:
@ValentinBuira bigger part of the code was actually moved from plugin to the mergin-client, it is already merged there [https://github.com/MerginMaps/python-api-client/pull/243](https://github.com/MerginMaps/python-api-client/pull/243).

PR #715, opened 2025-06-17
Fix crash when adding map sketches on older QGIS release and add a test case

ValentinBuira:
Fix crash when adding a map sketches layer because we can't use the new fancy enum in older QGIS release.

Also add a test case for this

image

### DIFF
--- a/Mergin/test/test_utils.py
+++ b/Mergin/test/test_utils.py
@@ -216,7 +216,8 @@ class test_utils(unittest.TestCase):
             dir_name, file_name = os.path.split(file_path)
             self.assertEqual(file_name, "map_sketches.gpkg")
 
-            layer = QgsVectorLayer(file_path, "", "ogr")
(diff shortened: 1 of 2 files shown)


JanCaha:
Not sure if we want to default to older enum type by default or if we want have something like:

```
 if Qgis.versionInt() >= 34000:
        symbol_layer.setDataDefinedProperty(QgsSymbolLayer.Property.StrokeColor, QgsProperty.fromExpression('"color"'))
    else:
        symbol_layer.setDataDefinedProperty(QgsSymbolLayer.PropertyStrokeColor, QgsProperty.fromExpression('"color"'))
```

This should work with both older versions of QGIS and also in the future if the `QgsSymbolLayer.PropertyStrokeColor` gets ever removed. Not sure if we have guideline for something like this. 

Besides that, it looks all good.

JanCaha (approved):

JanCaha:
Not exactly sure, why the test is failing. It should work the code seems to be fine ...

PR #742, opened 2025-07-02
Fix extent  in history viewer on unix

### LINKED ISSUE #701: Position of changes in history viewer
Problem:
In the map window of the History viewer, I don't see the position of the changes. The map zooms to the Null island.

To reproduce:
I experienced this in this project when I added more layers in one geopackage, but not sure if it is the core issue.

https://github.com/user-attachments/assets/0dbc180d-50ac-48e3-9baa-a4050a45bd00

ValentinBuira:
Ugly fix to make the extent work on macos and linux

the previous fix only worked on windows but in the end, the good behavior should be the one on unix, remove once we drop support for 3.34

Fix #701

### DIFF
--- a/Mergin/version_viewer_dialog.py
+++ b/Mergin/version_viewer_dialog.py
@@ -393,7 +393,7 @@ class VersionViewerDialog(QDialog):
             self.pan_tool = QgsMapToolPan(self.map_canvas)
             self.map_canvas.setMapTool(self.pan_tool)
 
-            self.current_diff = None
+            self.current_diff: QgsVectorLayer = None
             self.diff_layers = []
             self.filter_model = None
             self.layer_list.currentRowChanged.connect(self.diff_layer_changed)
@@ -590,6 +590,9 @@ class VersionViewerDialog(QDialog):
             self.map_canvas.setDestinationCrs(QgsProject.instance().crs())
             if layers:
                 extent = layers[0].extent()
+                print("layers[0].name()", layers[0].name())
+                print("layers[0].extent()", layers[0].extent())
+
                 d = min(extent.width(), extent.height())
                 if d == 0:
                     d = 1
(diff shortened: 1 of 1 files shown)


Mergin/version_viewer_dialog.py
```diff
@@ -590,17 +590,34 @@ def update_canvas(self, layers, set_extent=True):
             self.map_canvas.setDestinationCrs(QgsProject.instance().crs())
             if layers:
                 extent = layers[0].extent()
+                print("layers[0].name()", layers[0].name())
```
JanCaha, line 593: these prints should be removed - likely leftover, right?

Mergin/version_viewer_dialog.py
```diff
@@ -590,17 +590,34 @@ def update_canvas(self, layers, set_extent=True):
...
+                    # TODO bug specific on windows and older QGIS version
+                    # remove this madness and only keep above once you drop support for <=QGIS 3.34
+                    extent = (
+                        self.map_canvas.mapSettings().layerExtentToOutputExtent(layers[0], extent)
+                        if not layers[0].extent().isEmpty()
+                        else extent
+                    )
                 self.map_canvas.setExtent(extent)
 
+        # if self.current_diff:
```
JanCaha, line 613: remove commented code

Mergin/version_viewer_dialog.py
```diff
@@ -711,6 +728,8 @@ def add_all_to_project(self):
 
     def zoom_full(self):
         if self.current_diff:
+            print("self.current_diff.name()", self.current_diff.name())
```
JanCaha, line 731: again remove prints

Mergin/version_viewer_dialog.py
ValentinBuira: Thanks, removed

JanCaha:
Cannot replicate this on my setup, but the fix looks sensible

PR #747, opened 2025-07-03
Catch error in custom url

JanCaha:
catch error that occur on macOS in python 3.9 with invalid adress like ".something.com"

Based on @RastoHu report of the error:

screenshot_2025-07-02_at_13 53 49


ValentinBuira (approved):
Wasn't able to reproduce with my python version, but looks good to me

PR #758, opened 2025-07-09
better handling of not supporting sso on server

### LINKED ISSUE #753: SSO: Unexpected alerts when you type url address of custom server to plugin
**Steps to reproduce**

1. Remove everything from custom mergin maps server field
2. start typing http://locahost
3. Alerts re occurring after some ms

Image

JanCaha:
Fixes #753

### DIFF
--- a/Mergin/configuration_dialog.py
+++ b/Mergin/configuration_dialog.py
@@ -90,7 +90,7 @@ class ConfigurationDialog(QDialog):
         self.sso_email_required = False
 
         self.ui.button_sign_sso.clicked.connect(self.show_sign_sso)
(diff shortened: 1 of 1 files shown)


Mergin/configuration_dialog.py
```diff
@@ -191,8 +195,8 @@ def allow_sso_login(self) -> None:
             return
 
         allowed, msg = sso_login_allowed(self.server_url())
-        if msg:
-            QMessageBox.critical(self, "SSO allowed check", msg)
+        if not allowed:
+            self.show_sign_email()
```
uclaros, line 199: in `show_sign_email()` method, do we not also need to disable the sso button?

Issue #749, opened 2025-07-03
Unable to Delete Project Locally on Windows - Python Error


RastoHu:
**Description**
When attempting to delete a Mergin Maps project locally on Windows, users receive a Python error. Although the project disappears from the project tree in the Mergin Maps plugin UI, the project files remain in the local directory. This issue appears to be Windows-specific and does not occur on macOS.

**Environment**
Staging

**Application (+ app version, build, operating system)**
- QGIS version: 3.40.5-Bratislava
- Plugin version: 2025.3
- Windows: Windows 11

**Steps to reproduce**
1. Open QGIS with the Mergin Maps plugin on a Windows system
2. Navigate to a Mergin Maps project in the browser panel
3. Create New basic project
4. Right-click on this new project and select "Remove locally"
5. Confirm the removal
6. Observe the error message and check if files remain in the local directory

**Actual results**
- Error appears: "Warning Python error: An error has occurred while executing Python code"
- The project is removed from the Mergin Maps plugin project tree in the browser
- However, the project files still remain in the local directory

**Expected results**
The project should be properly deleted from both the plugin's project tree AND the local directory without any error messages.

**Screenshots or logs**
[No screenshots provided]

*Conclusion*
*Created by GitHub Copilot and supervised by a human test analyst.*

ValentinBuira:
Crap, the  

> QTimer.singleShot(250, lambda: shutil.rmtree(self.path))

is supposed to prevent this error

RastoHu:
tested on plugin build b111111 (https://github.com/MerginMaps/qgis-plugin/actions/runs/17461571338/artifacts/3926062179) and got error:
```
2025-09-08T12:05:21     WARNING    Traceback (most recent call last):
              File "C:\Users/hudak/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\Mergin\data_item.py", line 214, in <lambda>
              QTimer.singleShot(500, lambda: shutil.rmtree(self.path))
              ^^^^^^^^^^^^^^^^^^^^^^^^
              File "C:\PROGRA~1\QGIS34~1.10\apps\Python312\Lib\shutil.py", line 781, in rmtree
              return _rmtree_unsafe(path, onexc)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
              File "C:\PROGRA~1\QGIS34~1.10\apps\Python312\Lib\shutil.py", line 635, in _rmtree_unsafe
[...]

PR #789, opened 2025-08-27
fix on bug  "Unable to Delete Project Locally on Windows - Python Err…

DanChov:
fix python error. #749

On Windows, log file handlers may keep a lock on client-log.txt for a short
time even after being closed. This caused shutil.rmtree to fail with
PermissionError. Increased the QTimer delay before removal to give the OS
enough time to release the file lock.

delay 250 -> 500


varmar05:
Ok, let's keep it simple for time being - if problem persists we will need more robust cleaning strategy.

PR #796, opened 2025-09-01
Make photo sketching available for public use

tomasMizera:
- Made photo sketching available for public use
- Remove the whole "experimental" functionality as it had a few issues + we do not need it now
- Updated texts for photo sketching


MarcelGeo (approved):
It's still in preview? Looks good

PR #799, opened 2025-09-01
Adds better unhandled exception dialog during project sync

tomasMizera:
Enhanced the "unhandled exception" dialog:
- sending diagnostic log did not work before (passed `username` instead of the client ref)
- does not redirect users to GH issue tracker, but our support instead
- submits diagnostic log and opens email client with `mailto:` in one click
  - `mailto` message is preformatted, includes recipient, subject, body with the error message and a placeholder for user to enter more details about what happened
- adds optional checkbox if one does not want to share the log
- contains more UX-friendly messages

Hopefully, this should redirect most of the reports to the support team so we do not need to forward it manually ourselves.

| Old | New |
|--------|--------|
| Screenshot 2025-09-01 at 17 00 38 | Screenshot 2025-09-01 at 16 58 29 |

### DIFF
--- a/Mergin/projects_manager.py
+++ b/Mergin/projects_manager.py
@@ -181,7 +181,8 @@ class MerginProjectsManager(object):
                 unhandled_exception_message(
                     dlg.exception_details(),
                     "Project sync",
-                    f"Failed to sync project {project_name} due to an unhandled exception.",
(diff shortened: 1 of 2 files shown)


varmar05 (approved):
Nice!

RastoHu:
Unable to raise expected error. Only this was we ware able to get for python error:

PR #805, opened 2025-09-09
Fix del python error #749

DanChov:
Close all logging.FileHandlers under the project path and call logging.shutdown() to ensure .mergin/client-log.txt is unlocked
Trigger the deletion via QTimer.singleShot(400ms) so handlers have time to fully close, and call _delete_or_retry() 

[#749](https://github.com/MerginMaps/qgis-plugin/issues/749)

### DIFF
--- a/Mergin/data_item.py
+++ b/Mergin/data_item.py
@@ -1,3 +1,8 @@
+# -*- coding: utf-8 -*-
+
+# GPLv3 license
+# Copyright Lutra Consulting Limited
+
 from math import floor
 
 try:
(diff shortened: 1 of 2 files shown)


harminius (changes requested):
Nice.
Not sure which are your changes.. Apply your changes on the latest version of the code. I.e. when creating a local branch, always checkout to the current dev branch (`git checkout dev-2025.4.0`) and update the remote changes (`git pull`) Only then you want to create your local branch (`git checkout -b <my_new_local_branch>`)

PR #806, opened 2025-09-10
Fix windows del err

DanChov:
changes after code review
changed name from _delete_or_retry -> _delete_handlers
use self. path insted of path

### DIFF
--- a/Mergin/data_item.py
+++ b/Mergin/data_item.py
@@ -11,6 +11,7 @@ except ImportError:
     from PyQt6 import sip
 import os
 import shutil
+import logging
 import posixpath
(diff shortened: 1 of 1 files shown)


Mergin/data_item.py
```diff
@@ -158,6 +159,43 @@ def sync_project(self):
...
+        except (OSError, ValueError, RuntimeError):
+            pass
+
+    def _delete_handlers(self):
+        """Delete project directory (no rename, no retries)."""
+        try:
+            shutil.rmtree(self.path)
+        except PermissionError as e:
+            # Optional: user-facing message when files are still locked
+            QMessageBox.critical(
```
MarcelGeo, line 195: @tomasMizera do we want such critical message for users in case that some files are not possible to delete?

PR #811, opened 2025-09-15
Local path warning

### LINKED ISSUE #622: Syncing warning about local path for attachments on GNU/Linux
A user reported they are warned each time about a local path for attachment on GNU/Linux machine even it is all configured properly with relative paths

mm_warning

harminius:
Fixes #622 

The warning should display only when `Relative to Deafault Path` is selected from the dropdown (not only when there is some default path value)
image

### DIFF
--- a/Mergin/validation.py
+++ b/Mergin/validation.py
@@ -263,8 +263,6 @@ class MerginProjectValidator(object):
                                 expr = QgsExpression(cfg["DefaultRoot"])
                                 if not expr.isValid():
                                     self.issues.append(SingleLayerWarning(lid, Warning.ATTACHMENT_EXPRESSION_PATH))
-                            else:
-                                self.issues.append(SingleLayerWarning(lid, Warning.ATTACHMENT_DEFAULT_PATH))
 
                             # using hyperlinks for document path is not allowed when
                             if "UseLink" in cfg:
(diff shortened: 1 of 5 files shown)


Mergin/validation.py
```diff
@@ -247,12 +247,12 @@ def check_attachment_widget(self):
                     # check for relative paths
                     if "RelativeStorage" in cfg and cfg["RelativeStorage"] == 0:
                         self.issues.append(SingleLayerWarning(lid, Warning.ATTACHMENT_ABSOLUTE_PATH))
-                    if "DefaultRoot" in cfg:
+                    if "DefaultRoot" in cfg and cfg["RelativeStorage"] == 2:
```
MarcelGeo, line 250: Relative storage is checked in :up: condition, or?

Mergin/validation.py
```diff
@@ -245,21 +247,23 @@ def check_attachment_widget(self):
...
-                        expr = QgsExpression(cfg["DefaultRoot"])
-                        if expr.isValid():
-                            self.issues.append(SingleLayerWarning(lid, Warning.ATTACHMENT_EXPRESSION_PATH))
-
-                        # using hyperlinks for document path is not allowed when
-                        if "UseLink" in cfg:
-                            self.issues.append(SingleLayerWarning(lid, Warning.ATTACHMENT_HYPERLINK))
+                    if "RelativeStorage" in cfg:
+                        if cfg["RelativeStorage"] == 0:
+                            self.issues.append(SingleLayerWarning(lid, Warning.ATTACHMENT_ABSOLUTE_PATH))
```
tomasMizera, line 252: This one is questionable (not in this PR) - the app now accepts absolute paths in the form of network URL addresses. We should update the text of the validation.

Other conditions look good from the quick look, but must be properly tested (@RastoHu)

tomasMizera (approved):

MarcelGeo (changes requested):
Nice, I still do not understand why we upgraded tests to real geopackage? I think validations should run also with temp layer.

PR #818, opened 2025-09-25
Restart sync process

### LINKED ISSUE #774: Integrate restart of sync process
- acceptance criteria from parent
- try after 5 seconds sync processs if push to versions fails

MarcelGeo:
Resolves #774 

See relate workflow in https://github.com/MerginMaps/python-api-client/pull/262

In case of conflicted sync, we are trying to restart process after 5 sec for 10 times

image

:exclamation: 
- we could change branch after release 0.11.0 of py client and 4.0 of plugin
- update texts in dialogues 
- compatible with py client with get_push_changes_batch function

### DIFF
--- a/Mergin/projects_manager.py
+++ b/Mergin/projects_manager.py
@@ -5,6 +5,7 @@ import os
 from urllib.parse import urlparse
 from pathlib import Path
 import posixpath
+import json
 
 from qgis.core import QgsProject, Qgis, QgsApplication
 from qgis.utils import iface, OverrideCursor
@@ -29,6 +30,7 @@ from .utils import (
     login_error_message,
     same_dir,
     send_logs,
+    storage_limit_fail,
     unhandled_exception_message,
     unsaved_project_check,
     UnsavedChangesStrategy,
@@ -37,6 +39,7 @@ from .utils import (
     get_push_changes_batch,
     SYNC_ATTEMPTS,
     SYNC_ATTEMPT_WAIT,
+    push_error_message,
 )
 from .utils_auth import get_stored_mergin_server_url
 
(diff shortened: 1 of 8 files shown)


Mergin/projects_manager.py
```diff
@@ -366,102 +369,105 @@ def sync_project(self, project_dir, project_name=None):
...
-            if isinstance(dlg.exception, LoginError):
-                login_error_message(dlg.exception)
-            elif isinstance(dlg.exception, ClientError):
-                QMessageBox.critical(None, "Project sync", "Client error: " + str(dlg.exception))
-            elif isinstance(dlg.exception, AuthTokenExpiredError):
-                self.plugin.auth_token_expired()
+        has_push_changes = True
+        error_retries_attempts = 0
+        while has_push_changes:
+            dlg = SyncDialog()
```
varmar05, line 375: Do we need to create dlg in while loop?

Mergin/projects_manager.py
```diff
@@ -366,102 +369,105 @@ def sync_project(self, project_dir, project_name=None):
...
+            pull_timeout = 250
+            if error_retries_attempts > 0:
+                pull_timeout = SYNC_ATTEMPT_WAIT * 1000
+                dlg.labelStatus.setText("A sync conflict was detected. We are now retrying the synchronization to ensure your project is up to date.")
```
varmar05, line 379: The message seems a bit scary - would users know what `sync conflict` means?

Mergin/projects_manager.py
```diff
@@ -366,102 +369,105 @@ def sync_project(self, project_dir, project_name=None):
...
-                    # To note we check for a string since error in flask doesn't return server error code
-                    msg = "Somebody else is syncing, please try again later"
-                elif dlg.exception.server_code == ErrorCode.StorageLimitHit.value:
-                    msg = f"{dlg.exception.detail}\nCurrent limit: {bytes_to_human_size(dlg.exception.server_response['storage_limit'])}"
+            dlg = SyncDialog()
+            dlg.labelStatus.setText("Preparing project upload...")
+            dlg.push_start(self.mc, project_dir, project_name)
+            dlg.exec()  # blocks until success, failure or cancellation
+
+            qgis_proj_filename = os.path.normpath(QgsProject.instance().fileName())
```
varmar05, line 423: Could we add some comment here why we need this qgis file exercise here?

Mergin/projects_manager.py
MarcelGeo: Yes, It was like placeholder to something what we can discuss :)

Mergin/projects_manager.py
MarcelGeo: yeah, this was here before ...

Issue #678, opened 2025-03-14
Inform users about use of forbidden characters


gioman:
I have setup a project and i MM properties set up the photo name as

'media/tampa_' || now()

which returns something like

![Image](https://github.com/user-attachments/assets/983efba5-9150-48a0-a716-2cd730a0d88e)

This is ok on the mobile that syncs no problem, but the plugin spits out this python error:

```
Traceback (most recent call last):

  File "C:\Users/giovanni/AppData/Roaming/QGIS/QGIS3\profiles\test/python/plugins\Mergin\sync_dialog.py", line 275, in pull_timer_tick
    is_running = pull_project_is_running(self.job)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\Users/giovanni/AppData/Roaming/QGIS/QGIS3\profiles\test/python/plugins\Mergin\mergin\client_pull.py", line 527, in pull_project_is_running
    raise future.exception()

  File "C:\OSGeo4W\apps\Python312\Lib\concurrent\futures\thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "C:\Users/giovanni/AppData/Roaming/QGIS/QGIS3\profiles\test/python/plugins\Mergin\mergin\client_pull.py", line 95, in _do_download
    item.download_blocking(mc, mp, project_path)

  File "C:\Users/giovanni/AppData/Roaming/QGIS/QGIS3\profiles\test/python/plugins\Mergin\mergin\client_pull.py", line 324, in download_blocking
    save_to_file(resp, self.download_file_path)

  File "C:\Users/giovanni/AppData/Roaming/QGIS/QGIS3\profiles\test/python/plugins\Mergin\mergin\utils.py", line 39, in save_to_file
    with open(path, "wb") as output:
         ^^^^^^^^^^^^^^^^

OSError: [Errno 22] Invalid argument: 'C:\\Users\\giovanni\\Documents\\test\\.mergin\\fetch_v10-v11\\media\\tampa_2025-03-13T20:53:24.694.jpg.0'
```

Maybe would be better to not allow characters like "-" or "/" in the expression to set the photo path and name.

ValentinBuira:
Hi @gioman, you should not be able to sync even on mobile

see bottom warning message
[...]

PR #819, opened 2025-09-29
require hardcoded expressions

harminius:
fixes https://github.com/MerginMaps/qgis-plugin/issues/678#issuecomment-3346619834

### DIFF
--- a/Mergin/test/test_validations.py
+++ b/Mergin/test/test_validations.py
@@ -106,6 +106,14 @@ class test_validations(unittest.TestCase):
         validator.check_attachment_widget()
         self.assertTrue(len(validator.issues) == 0)
 
+        # bad expression
+        config["DefaultRoot"] = "now()"
+        widget_setup = QgsEditorWidgetSetup("ExternalResource", config)
+        self.mem_layer.setEditorWidgetSetup(photo_field_idx, widget_setup)
+        validator.check_attachment_widget()
+        self.assertTrue(len(validator.issues) == 1)
+        validator.issues = []
+
         # relative to project path
         config["RelativeStorage"] = 1
         config.pop("DefaultRoot", None)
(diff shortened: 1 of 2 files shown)


Mergin/utils.py
```diff
@@ -1679,3 +1683,33 @@ def is_inside(proj_dir: str, to_check: str) -> bool:
...
+    if val is None:
+        return ""
+
+    if isinstance(val, str):
+        return val.strip()
+
+    if isinstance(val, QDateTime):
+        return val.toString()
+
+    if isinstance(val, QgsGeometry):
```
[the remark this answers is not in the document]
harminius: That doesn't work, unfortunately

```
print(str(val))
PyQt5.QtCore.QDateTime(2025, 9, 30, 9, 51, 46, 417)
print(val.toString())
Tue Sept 30 09:51:46 2025
```

Mergin/project_settings_widget.py
```diff
@@ -137,6 +138,7 @@ def __init__(self, parent=None):
         else:
             # Hide by default
             self.groupBox_photo_sketching.setVisible(False)
+        # self.label_preview.setPlainText()
```
tomasMizera, line 141: Commented out code

Mergin/utils.py
```diff
@@ -1679,3 +1683,33 @@ def is_inside(proj_dir: str, to_check: str) -> bool:
         return True
     except ValueError:
         return False
+
+
+def val_to_string(val: Any) -> str:
```
tomasMizera, line 1688: I believe `val` in this case is Qt's [QVariant](https://doc.qt.io/qt-6/qvariant.html) 
```suggestion
def qvariant_to_string(val: Any) -> str:
```
otherwise, if it can be anything, you can go with simple

```suggestion
def to_string(val: Any) -> str:
```

harminius:
> @harminius we do not do what you say in the first line
> 
> > 1, The easiest is to require specific expressions for attachment location.
> 
> right?

That was meant to [this part](https://github.com/MerginMaps/qgis-plugin/pull/819/files#diff-a135ae715e10f93b23e3891cca47086b8fce4bade3c44d873a589b0681978adfL265) - in attachment widget validations, we don't need to evaluate the expression and check if it is inside the project folder
In photo name setting in order to know what the final value is, we need to evaluate the expression.

PR #833, opened 2025-11-08
Fix loading on Qt6 QGIS

ViperMiniQ:
Remove Qt5 import.


harminius (approved):
Yeah, thanks for fixing my fault.

PR #836, opened 2025-11-14
Reset changes fail on windows

DanChov:
Fix for a Windows-specific issue where GPKG files remain locked briefly after closing the project, which causes make_copy_sqlite to fail during reset_local_changes().
The solution reuses the same approach already implemented in the plugin for the unfinished pull workflow: after clearing the project, a short delay allows Windows to release SQLite/OGR file handles before the reset is executed. This aligns the reset logic with the proven, existing behavior used elsewhere in the plugin.
image
image
image

### DIFF
--- a/Mergin/create_project_wizard.py
+++ b/Mergin/create_project_wizard.py
@@ -170,6 +170,7 @@ class ProjectSettingsPage(ui_proj_settings, base_proj_settings):
         if not self.project_workspace_cbo.currentData(Qt.ItemDataRole.UserRole):
             self.create_warning("You do not have permissions to create a project in this workspace!")
             return
+
         proj_name = self.project_name_ledit.text().strip()
         if not proj_name:
             self.create_warning("Project name missing!")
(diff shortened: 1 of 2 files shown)


Mergin/projects_manager.py
```diff
@@ -308,32 +308,48 @@ def reset_local_changes(self, project_dir: str, files_to_reset=None):
...
         current_project_path = os.path.normpath(QgsProject.instance().absolutePath())
+
+        # Windows-specific behavior:
+        # When a project is opened from this directory, QGIS may keep GPKG file handles
+        # for a short time after closing the project. The same workaround is used in
+        # `close_project_and_fix_pull()` (unfinished pull handling).
+        delay = 0
         if current_project_path == os.path.normpath(project_dir):
             QgsProject.instance().clear()
+            QApplication.processEvents()
```
harminius: what's doing this? 
i don't see it used elsewhere..

Issue #856, opened 2026-01-21
Relations warnings - Better logging output


fernandinand:
For duplicates values from relations sync warning, we should have some meaningful output, namely:

* The target relation key field (or list in case of multiple fields)
* At least a sample of duplicated value(s) to ease debug

<img width="732" height="306" alt="Image" src="https://github.com/user-attachments/assets/562907c3-2970-43c6-bddb-480113a0a2bb" />

PR #864, opened 2026-02-10
Improve relation key uniqueness warning with field name and samples

DanChov:
Fixes: https://github.com/MerginMaps/qgis-plugin/issues/856

This PR improves the relation key uniqueness warning by showing which field contains duplicated values and a small sample of those values.

Previously, the warning only said that duplicates exist, which made it hard to find the problem in the layer. Now the message includes the field name and example duplicated values to make debugging easier.

The number of sample values is controlled by a single limit in the code and can be easily adjusted if needed. The change keeps the existing warning system and does not affect other validations.

image

### DIFF
--- a/.github/workflows/notify-milestone.yml
+++ b/.github/workflows/notify-milestone.yml
@@ -1,0 +1,20 @@
+name: Notify releases repo
+
+on:
+  milestone:
+    types: [created, edited]
+
+jobs:
+  notify:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Dispatch to releases repo
+        env:
+          GH_TOKEN: ${{ secrets.MM_RELEASES_SYNC_TOKEN }}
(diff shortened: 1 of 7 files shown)


Mergin/validation.py
```diff
@@ -484,7 +514,25 @@ def warning_display_string(warning_id, url=None):
     elif warning_id == Warning.DATABASE_SCHEMA_CHANGE:
         return "Database schema was changed"
     elif warning_id == Warning.KEY_FIELD_NOT_UNIQUE:
-        return "Relation key field contains duplicated values"
+        # if no details passed, keep old message
+        # if details passed (dict with type), enrich message with field(s) + sample values
```
harminius: don't get this..
This warning comes only from the edited (enriched) source, right? 
How can it come without details?

PR #870, opened 2026-02-13
Respect the Qgis SSL configuration on trusted certificates and trusted root CA configured in settings

### LINKED ISSUE #820: Use certificates from the operating system
In some cases, users may get this kind of errors when trying to connect to Mergin Maps server:

```
[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1010)
```

This could be because the corporate network as an appliance that does SSL inspection (man-in-the-middle proxy), and data sent to the MM plugin are signed by their own internal certificate. This certificate may be stored trusted by the operating system, but python's `requests` module (that we use for all networking) does not use trusted certificates from the OS - it rather uses `certifi` module (which has its bundle of root certificates).

To avoid this problem, we could use `truststore` module (https://github.com/sethmlarson/truststore) that should be able to inject trusted certificates from the OS. It is a pure python module, so it should be easy to add as a dependency. The only(?) issue is that python 3.10+ is required.

lleirborras:
Also, fixes  https://github.com/MerginMaps/qgis-plugin/issues/820

### DIFF
--- a/Mergin/utils_auth.py
+++ b/Mergin/utils_auth.py
@@ -4,12 +4,11 @@
 import hashlib
 import os
 import re
+import sys
 import typing
 import uuid
 import json
 from urllib.error import URLError
-import requests
-import urllib3
 from enum import Enum
 
 from qgis.core import (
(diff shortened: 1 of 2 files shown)


Mergin/utils_auth.py
```diff
@@ -535,14 +535,65 @@ def mergin_server_deprecated_version(url: str) -> bool:
...
+        plugin_dir = os.path.dirname(os.path.realpath(__file__))
+        bundled_cert = os.path.join(plugin_dir, "mergin", "cert.pem")
+        if os.path.exists(bundled_cert):
+            marker = "# --- QGIS trusted CAs ---"
+            with open(bundled_cert, "r") as f:
+                existing = f.read()
+            if marker not in existing:
+                with open(bundled_cert, "a") as f:
+                    f.write(f"\n{marker}\n")
+                    f.write(qgis_ca_pem)
```
wonder-sk: I would prefer not to have existing files in the plugin's directory modified. Additionally, there is an issue that if the QGIS trusted CAs have been written once, they won't get updated if some certs are manually added/removed later.

My preference would be this: in the `mergin` python module, we should add a new module-wide function - e.g. mergin.set_trusted_certificates() which could have a file path to certificates file as the only argument. Then, whenever `MerginClient` is created, its internal urllib configuration would be set to use that (and in case there is no certificates file provided, it would fall back to the existing CA cert handling https://github.com/MerginMaps/python-api-client/blob/master/mergin/client.py#L174). Hopefully that should be cleaner, without two different levels of mac-specific handling... What do you think?

wonder-sk (commented):
Thanks for your contribution!

Nice idea to use QGIS trusted CA certs - I thought QGIS API provided only the custom ones added by the user, but it actually returns all of them, so we don't need to deal with `truststore` module...

wonder-sk:
Sorry for the slow response. There are still some issues with this PR, e.g. `set_trusted_certificates()` not being present in the mergin client. We have a better fix coming in the python api client repo - https://github.com/MerginMaps/python-api-client/pull/306 - so I will close this PR

PR #880, opened 2026-02-24
ci: mimic QGIS repository checks with bandit and flake8

DanChov:
https://github.com/MerginMaps/server-private/issues/3229


harminius (commented):
If it is tried locally, I would give it a try. 🟢 

I think ruff should also know about the line-length..`--line-length 120`. Does it fix the issues without `--fix` flag?

PR #896, opened 2026-03-25
Update tests and test running on CI

JanCaha:
Update tests to pytests and update infrastructure on CI for running tests


MarcelGeo (changes requested):
Some files are just moved (?). This approach is more useful when using cache - nice.

MarcelGeo (approved):

PR #875, opened 2026-02-19
Qgis security scan

harminius:
Resolves: https://github.com/MerginMaps/qgis-plugin/issues/872

Includes:
- Security fixes from bandit  [commits 2-7](cd2c4a75deb3e6a8f6b24d49c29573b74594a8f3...df55ae1eb32a598507a829293577f42c51f3fad3)
- fix for detect-secrets check: [commit 8](107e58855dea12b53774ba320653d32edf1031ba)
- Code Quality fix from ruff: [commit 9 ](3a6df3ae53ea4ba337e1ed1e2b73817c7830d726) 
- Manual Code Quality fix of issue from flake8: [commit 10 ](91deeaee50202bf96adf01c6f6263ca5ffb4d5e0)

[QGIS docs reference](https://plugins.qgis.org/docs/security-scanning)
[QGIS check - code](https://github.com/qgis/QGIS-Plugins-Website/blob/master/qgis-app/plugins/security_scanner.py#L91)

Related to https://github.com/MerginMaps/python-api-client/pull/293

Issue #473, opened 2022-11-25
Plugin database is locked error


jozef-budac:
During testing of input on phone with full storage, this error appears in plugin.
Probably input removes some data or packages, and plugin didn't like it ;)

```
An error has occurred while executing Python code: 

sqlite3.OperationalError: database is locked 
Traceback (most recent call last):
  File "/home/jozef/.local/share/QGIS/QGIS3/profiles/prof1/python/plugins/Mergin/plugin.py", line 390, in current_project_sync
    self.manager.project_status(self.mergin_proj_dir)
  File "/home/jozef/.local/share/QGIS/QGIS3/profiles/prof1/python/plugins/Mergin/projects_manager.py", line 168, in project_status
    pull_changes, push_changes, push_changes_summary = self.mc.project_status(project_dir)
  File "/home/jozef/.local/share/QGIS/QGIS3/profiles/prof1/python/plugins/Mergin/mergin/client.py", line 779, in project_status
    push_changes = mp.get_push_changes()
  File "/home/jozef/.local/share/QGIS/QGIS3/profiles/prof1/python/plugins/Mergin/mergin/merginproject.py", line 327, in get_push_changes
    size, checksum = do_sqlite_checkpoint(self.fpath(file["path"]), self.log)
  File "/home/jozef/.local/share/QGIS/QGIS3/profiles/prof1/python/plugins/Mergin/mergin/utils.py", line 92, in do_sqlite_checkpoint
    cursor.execute("PRAGMA wal_checkpoint=FULL")
sqlite3.OperationalError: database is locked


Python version: 3.10.6 (main, Nov  2 2022, 18:53:38) [GCC 11.3.0] 
QGIS version: 3.22.13-Białowieża Białowieża, 02ff117c4e 

Python Path:
/usr/share/qgis/python
/home/jozef/.local/share/QGIS/QGIS3/profiles/prof1/python
/home/jozef/.local/share/QGIS/QGIS3/profiles/prof1/python/plugins
/usr/share/qgis/python/plugins
/usr/lib/python310.zip
/usr/lib/python3.10
/usr/lib/python3.10/lib-dynload
/home/jozef/.local/lib/python3.10/site-packages
/usr/local/lib/python3.10/dist-packages
/usr/lib/python3/dist-packages
/home/jozef/.local/share/QGIS/QGIS3/profiles/prof1/python
.
/home/jozef/Lutra/projects/b/Aaa

PR #907, opened 2026-05-11
Fix sqlite3 cursor and placeholder syntax in get_row_from_db (#895)

harminius:
## Summary

  Fixes #895 — clicking a version in the Changes viewer raised:

  ```
  TypeError: 'sqlite3.Cursor' object does not support the context manager protocol
    at Mergin/diff.py:78 in get_row_from_db
  ```

  `get_row_from_db` in `Mergin/diff.py` was written with psycopg2 conventions even though `db_conn` is a stdlib `sqlite3.Connection`:

  - `with db_conn.cursor() as c:` — `sqlite3.Cursor` is not a context manager (only `sqlite3.Connection` is).
  - `'"{}" = %s'` placeholders — `sqlite3` uses `?`, not `%s`.

  ## Changes

  - Drop the `with` wrapper; assign cursor directly
  - `%s` → `?` for parameter placeholders.

### DIFF
--- a/Mergin/diff.py
+++ b/Mergin/diff.py
@@ -75,7 +75,6 @@ def get_row_from_db(db_conn, schema_table, entry_changes):
     Fetches a single row from DB's table based on the values of pkeys
     in changeset entry
     """
-    c = db_conn.cursor()
     where_clauses = []
     query_params = []
 
@@ -88,9 +87,7 @@ def get_row_from_db(db_conn, schema_table, entry_changes):
     # We parameterize values securely; table/column names are trusted internal schema objects.
     query = 'SELECT * FROM "{}" WHERE {}'.format(schema_table.name, " AND ".join(where_clauses))  # nosec B608
 
-    c.execute(query, tuple(query_params))
-
-    return c.fetchone()
+    return db_conn.execute(query, tuple(query_params)).fetchone()
 
 
 def parse_gpkg_geom_encoding(wkb_with_gpkg_hdr):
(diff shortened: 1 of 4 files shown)


varmar05:
@harminius does this mean this was never working?

also I do not see using context manager with connection object - would be good to review gpkg handling here and wrap it up

harminius:
@varmar05 
The bug(s) have been there for two months, as it was introduced with SQL sanitization https://github.com/MerginMaps/qgis-plugin/pull/875/changes/cd2c4a75deb3e6a8f6b24d49c29573b74594a8f3

As of context manages
- `with` works unintuintively - it's transaction-based, not closing the connection https://blog.rtwilson.com/a-python-sqlite3-context-manager-gotcha/
- we can use simple try/finally
- or there is built-in `contextlib` with `closing` function if you prefer  https://docs.python.org/3/library/contextlib.html

harminius:
unlike the `connection`, `cursor` doesn't lock any OS-level resources
sqlite3 can create the cursor implicitly with the `.execute()` method

I would prefer not to worry about its lifecycle and use the shortcut

https://docs.python.org/3/library/sqlite3.html#sqlite3-connection-shortcuts

varmar05 (approved):

harminius:
Closing the SQL connection might fix this dinosaur: https://github.com/MerginMaps/qgis-plugin/issues/473

MarcelGeo (approved):

PR #916, opened 2026-05-26
New SQL for multi_select

### LINKED ISSUE #914: FILTERING: Update the template query for Value Relation with AllowMulti
Update the plugin implementation of the filtering template query for Value Relation with AllowMulti to something like this:
`"sql_expression": "(',' || TRIM(\"NAME_OF_FIELD\", '{}') || ',' ) LIKE '%,' || @@value@@ || ',%'"`

JanCaha:
Fixes #914

### DIFF
--- a/Mergin/field_filtering.py
+++ b/Mergin/field_filtering.py
@@ -112,6 +112,7 @@ class FieldFilter:
         f.filter_type = FieldFilterType(data["filter_type"])
         f.filter_name = data["filter_name"]
         f.sql_expression = data.get("sql_expression", "")
+        f.field_has_multi_selection = data.get("field_has_multi_selection", False)
         if not f.sql_expression:
             f._generate_sql_expression()
         return f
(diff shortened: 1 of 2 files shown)


Mergin/field_filtering.py
```diff
@@ -86,10 +85,14 @@ def __init__(
...
         self.layer_id = ""
 
+        allow_multi = None
         if layer is not None:
             provider = layer.dataProvider()
             self.provider = provider.name() if provider else ""
             self.layer_id = layer.id()
+            allow_multi = layer.fields().field(field_name).editorWidgetSetup().config().get("AllowMulti", None)
+
+        self.field_has_multi_selection = allow_multi is True
```
MarcelGeo, line 95: Just wondering about that new attribute. This could not fail in `from_dict` and `to_dict` ? As there is 

```python
if not f.sql_expression:
     f._generate_sql_expression()
```

MarcelGeo (changes requested):
As we have there tests for this field filtering, we could introduce some of them for such case.

PR #925, opened 2026-06-23
Fix CI tests failure

JanCaha:
seems like qca is not install with qgis 3.44


MarcelGeo:
Looks like failing now @JanCaha

